PR #264, opened 2024-09-18
MTG-374 Make consistency checks on API optional

n00m4d:
# What
This PR makes consistency checks on API side optional.
# Why
In devnen we cannot backfill whole the transactions history and because of that API does not return any info about some assets which are from incosistent trees, even simple getAsset may not work.
# How
Make a few config parameters optional. Now we can even turn on only one consistency checks(if we ever need it)


RequescoS (approved):

StanChe (approved):
I believe we have docs somewhere outlining the env vars, please do update those as well

PR #273, opened 2024-09-30
Added code doc for `dump_client.rs`

rwwwx:


n00m4d (approved):

rocks-db/src/dump_client.rs
```diff
@@ -83,6 +83,16 @@ struct AssetRecord {
 }
 
 impl Storage {
+    /// It concurrently creates a `CSV file` for each provided file.
+    /// This separation into files makes sense because that each file corresponds to a separate table in `Postgres`.
+    ///
+    /// # Args:
+    /// `metadata_file_and_path` - The file and path corresponding to the metadata data.
```
RequescoS, line 90: I think it would be better to indicate which table in Postgre data corresponds to. For example metadata corresponds to 'tasks' table

rocks-db/src/dump_client.rs
```diff
@@ -168,30 +171,30 @@ impl Storage {
...
         for k in iter
             .filter_map(|k| k.ok())
             .filter_map(|(key, _)| decode_pubkey(key.to_vec()).ok())
         {
             batch.push(k);
+
+            // When batch is filled, find `AssetIndex` and send it to `tx_indexes` channel.
             if batch.len() == batch_size {
                 let indexes = self
                     .get_asset_indexes(batch.as_ref())
```
RequescoS, line 200: Cant we use std::mem::take(&mut batch). Then it will be no need to call batch.clear()

RequescoS (approved):

rocks-db/src/dump_client.rs
```diff
@@ -83,6 +83,16 @@ struct AssetRecord {
 }
 
 impl Storage {
+    /// It concurrently creates a `CSV file` for each provided file.
+    /// This separation into files makes sense because that each file corresponds to a separate table in `Postgres`.
```
StanChe, line 87:
```suggestion
    /// Concurrently dumps data into several `CSV files`, where each file corresponds to a separate table in the index database (`Postgres`).
```

StanChe (approved):

PR #277, opened 2024-10-02
[MTG-711] feat: showZeroBalance

RequescoS:


postgre-client/src/asset_filter_client.rs
```diff
@@ -227,6 +231,10 @@ fn add_filter_clause<'a>(
                     query_builder.push(" OR fungible_tokens.fbt_owner = ");
                     query_builder.push_bind(owner_address);
                     query_builder.push(" ) ");
+                    if !options.show_zero_balance {
```
StanChe, line 234: This should be inside the OR clause like `and (asset.owner = xxx or (ft.owner = xxx and ft.balance > 0))`

migrations/7_fungible_balance.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE fungible_tokens ADD COLUMN fbt_balance bigint NOT NULL DEFAULT 0;
+ALTER TABLE fungible_tokens ADD COLUMN fbt_slot_updated bigint NOT NULL DEFAULT 0;
+CREATE INDEX fungible_tokens_fbt_balance_idx ON fungible_tokens(fbt_balance) WHERE fbt_balance > 0;
```
StanChe, line 3: do we delete and recreate this index in the full sync? The idea to drop all the indexes is to ensure PG is not updating any indexes while loading from csv and only creates it once, when all the data is in

StanChe (changes requested):

migrations/7_fungible_balance.sql
RequescoS: Yeah, i forget to add new indexes into drop_indexes() function. Thank you!

StanChe (approved):

rwwwx (approved):

PR #278, opened 2024-10-03
feat: show_fungible option for GetBy methods

RequescoS:


nft_ingester/src/api/dapi/search_assets.rs
```diff
@@ -39,6 +39,9 @@ pub async fn search_assets<TPF: TokenPriceFetcher>(
     token_price_fetcher: Arc<TPF>,
     metrics: Arc<ApiMetricsConfig>,
 ) -> Result<AssetList, StorageError> {
+    if options.show_fungible {
```
StanChe, line 42: is it expected to replace the token_type filter value when the show_fungible value is set

StanChe (approved):
Looks ok, couple minor comments

nft_ingester/src/api/dapi/search_assets.rs
```diff
@@ -22,14 +22,14 @@ use super::asset_preview::populate_previews;
...
-    filter: SearchAssetsQuery,
+    mut filter: SearchAssetsQuery,
     sort_by: AssetSorting,
     limit: u64,
     page: Option<u64>,
     before: Option<String>,
     after: Option<String>,
     cursor: Option<String>,
-    options: SearchAssetsOptions,
+    options: GetByMethodsOptions,
```
n00m4d, line 32: Why do we introduce new struct `GetByMethodsOptions` and use it here instead of `SearchAssetsOptions`.
First time I thought it will be used only in `getBy...` requests but if use it for `search` query as well why not we just add one more field to `SearchAssetsOptions` ?
Is there any issues with that that I don't see?

postgre-client/src/asset_filter_client.rs
```diff
@@ -172,13 +172,13 @@ fn add_filter_clause<'a>(
         group_clause_required = true;
     }
     if let Some(ref token_type) = filter.token_type {
-        if token_type == &TokenType::All {
+        if token_type == &TokenType::All && filter.owner_address.is_some() {
```
n00m4d, line 175: Are we sure it will not break any `serch` queries? Because this function is used by that enpoint as well.
For example if someone will call `serch` and put `TokenType::Fungible` but will not point owner, what gonna happen?

RequescoS: There validation check in API that does not allow setting tokenType without setting ownerAddress

nft_ingester/src/api/dapi/search_assets.rs
RequescoS: SearchAssetsOptions does not need to contain the `show_fungible` field, while GetByMethodsOptions need to. We use json-rpc lib, which validates input parameters, so we use different structs for allowing setting different options for different methods

n00m4d (approved):

PR #281, opened 2024-10-10
More metrics for synchronizer

n00m4d:


rocks-db/src/dump_client.rs
```diff
@@ -435,20 +460,27 @@ impl Storage {
...
+                let start = Instant::now();
+
                 if let Err(e) = writer.serialize(k).map_err(|e| e.to_string()) {
                     error!(
                         "Error while writing data into {:?}. Err: {:?}",
                         file_and_path.1, e
                     );
                 }
+
+                synchronizer_metrics.set_file_write_time(start.elapsed().as_millis() as f64);
```
StanChe, line 472: I'd suggest using a counter of writes per file, or making it a histogram per file

rocks-db/src/dump_client.rs
```diff
@@ -275,6 +296,7 @@ impl Storage {
             if rx_indexes_cloned.is_empty() {
                 continue;
             } else if let Ok(indexes) = rx_indexes_cloned.try_recv() {
+                let start = Instant::now();
                 for (key, index) in indexes {
                     let metadata_url = index
                         .metadata_url
```
StanChe, line 302: I'd advise on increasing some metric on every .send to any channel (like the tx_metadata_cloned)

StanChe (approved):
I'd extend it a bit

PR #282, opened 2024-10-11
MTG-742 Move fungible assets dump to different thread

n00m4d:
# What
This PR moves dumping fungible assets, during full synchronization, to different blocking tokio thread.

# Why
To speed up full sync process and avoid complicated selects from RocksDB.

### DIFF
--- a/rocks-db/src/dump_client.rs
+++ b/rocks-db/src/dump_client.rs
@@ -176,8 +176,14 @@ impl Storage {
 
         // dump fungible assets in separate blocking thread
         let column: Column<TokenAccount> = Self::column(self.db.clone(), self.red_metrics.clone());
+        let cloned_metrics = synchronizer_metrics.clone();
         let fungible_assets_join = tokio::task::spawn_blocking(move || {
-            Self::dump_fungible_assets(column, fungible_tokens_file_and_path, batch_size)
+            Self::dump_fungible_assets(
+                column,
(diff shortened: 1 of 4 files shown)


rocks-db/src/dump_client.rs
```diff
@@ -264,6 +278,64 @@ impl Storage {
...
+            }
+            batch.clear();
+        }
+
+        if let Err(e) = writer.flush().map_err(|e| e.to_string()) {
+            error!(
+                "Error happened during flushing data to {:?}. Err: {:?}",
+                file_and_path.1, e
+            );
+        }
```
StanChe, line 336: Please add a log indicating that this process has finished working

rocks-db/src/dump_client.rs
```diff
@@ -264,6 +278,64 @@ impl Storage {
...
+            .filter_map(|(_, value)| deserialize::<TokenAccount>(value.to_vec().as_ref()).ok())
+        {
+            batch.push((
+                token.pubkey.to_string(),
+                token.owner.to_string(),
+                token.amount,
+                token.slot_updated,
+            ));
+
+            if batch.len() >= batch_size {
```
StanChe, line 306: Please add some metrics

StanChe (approved):
Looks great! Add some metrics (at least a counter) and a log. I'd also rely on @RequescoS to verify the right data is dumped

PR #283, opened 2024-10-11
MTG-747 Save updates for fungible assets during regular sync

n00m4d:
# What
This PR adds fungible assets sync during regular synchronization. Also from now on, for fungible assets, we put two rows to index column family. And it adds new row to Postgre - asset pubkey.

### DIFF
--- a/migrations/8_token_acc_key.sql
+++ b/migrations/8_token_acc_key.sql
@@ -1,12 +1,16 @@
-ALTER TABLE fungible_tokens ADD COLUMN fbt_pubkey bytea NOT NULL;
+DROP TABLE fungible_tokens;
 
-ALTER TABLE fungible_tokens DROP CONSTRAINT fungible_tokens_pkey;
+-- recreate table and add fbt_pubkey
 
-ALTER TABLE fungible_tokens ADD PRIMARY KEY (fbt_pubkey);
-
-DROP INDEX IF EXISTS fungible_tokens_fbt_balance_idx;
-
-DROP INDEX IF EXISTS fungible_tokens_fbt_slot_updated_idx;
+CREATE TABLE fungible_tokens (
+    fbt_pubkey bytea NOT NULL,
+    fbt_owner bytea NOT NULL,
+    -- mint key
+    fbt_asset bytea NOT NULL,
+    fbt_balance bigint NOT NULL DEFAULT 0,
+    fbt_slot_updated bigint NOT NULL DEFAULT 0,
+    PRIMARY KEY (fbt_pubkey)
+);
 
 CREATE INDEX fungible_tokens_fbt_owner_balance_idx ON fungible_tokens(fbt_owner, fbt_balance);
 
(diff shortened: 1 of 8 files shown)


migrations/8_token_acc_key.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE fungible_tokens ADD COLUMN fbt_pubkey bytea NOT NULL;
```
StanChe, line 1: Why not drop and recreate it?

rocks-db/src/batch_client.rs
```diff
@@ -256,6 +265,33 @@ impl AssetIndexReader for Storage {
             }
         }
 
+        // compare to other data this data is saved in HashMap by token account keys, not mints
+        for token_acc in token_accounts_details.iter().flatten() {
```
StanChe, line 269: Can you get multiple accounts with the same pubkey?

migrations/8_token_acc_key.sql
```diff
@@ -0,0 +1,13 @@
...
+
+ALTER TABLE fungible_tokens DROP CONSTRAINT fungible_tokens_pkey;
+
+ALTER TABLE fungible_tokens ADD PRIMARY KEY (fbt_pubkey);
+
+DROP INDEX IF EXISTS fungible_tokens_fbt_balance_idx;
+
+DROP INDEX IF EXISTS fungible_tokens_fbt_slot_updated_idx;
+
+CREATE INDEX fungible_tokens_fbt_owner_balance_idx ON fungible_tokens(fbt_owner, fbt_balance);
```
StanChe, line 11: Those indexes should be dropped and recreated in batch load in Postgres, can't see that in this pr

rocks-db/src/batch_client.rs
n00m4d: We cannot get multiple accounts with different data but same pubkey

migrations/8_token_acc_key.sql
RequescoS: yeah, we need to drop old table because on devnet it is already filled and such migration may fail

rocks-db/src/batch_client.rs
StanChe: And we'll also not get those token accounts from other details, so how can it be in asset_index?

postgre-client/src/asset_index_client.rs
```diff
@@ -213,20 +216,23 @@ pub(crate) fn split_assets_into_components(asset_indexes: &[AssetIndex]) -> Asse
...
-            i.fungible_tokens.iter().map(|fungible_token| {
-                (
-                    (fungible_token.owner, fungible_token.asset),
-                    *fungible_token,
-                )
-            })
-        })
-        .collect::<HashMap<_, _>>();
+    let mut fungible_tokens = vec![];
+    for asset in asset_indexes.iter() {
```
StanChe, line 220: can be just asset_indexes.filter_map

StanChe (approved):

PR #284, opened 2024-10-14
MTG-780 Optional collection auth map

n00m4d:
# What
This PR is one more synchronizer refactor work. It adds optional hashMap to get_asset_indexes method. That map contains collection key and authority value. It's necessary only for core assets.

# Why
It should decrease number of times we query DB during asset batch processing. For full synchronization process it can save us hours(potentially).

### DIFF
--- a/rocks-db/src/batch_client.rs
+++ b/rocks-db/src/batch_client.rs
@@ -103,7 +103,11 @@ impl AssetUpdateIndexStorage for Storage {
 
 #[async_trait]
 impl AssetIndexReader for Storage {
-    async fn get_asset_indexes(&self, keys: &[Pubkey]) -> Result<HashMap<Pubkey, AssetIndex>> {
+    async fn get_asset_indexes<'a>(
+        &self,
+        keys: &[Pubkey],
+        collection_authorities: Option<&'a HashMap<Pubkey, Pubkey>>,
+    ) -> Result<HashMap<Pubkey, AssetIndex>> {
         let mut asset_indexes = HashMap::new();
         let start_time = chrono::Utc::now();
         let assets_static_fut = self.asset_static_data.batch_get(keys.to_vec());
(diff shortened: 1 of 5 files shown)


rocks-db/src/batch_client.rs
```diff
@@ -142,13 +146,29 @@ impl AssetIndexReader for Storage {
             .iter()
             .flat_map(|c| c.as_ref().map(|c| c.collection.value))
             .collect::<Vec<_>>();
```
StanChe, line 148: You need to collect those pks only when you don't have the collection_authorities passed in.

rocks-db/src/batch_client.rs
```diff
@@ -142,13 +146,29 @@ impl AssetIndexReader for Storage {
...
+                    .into_iter()
+                    .flatten()
+                    .filter_map(|asset| asset.authority.value.map(|v| (asset.pubkey, v)))
+                    .collect::<HashMap<_, _>>()
+            }
+        };
+
+        // mpl_core_map.is_empty() && collection_authorities.is_some() check covers case when DB is empty
+        let mpl_core_collections = if mpl_core_map.is_empty() && collection_authorities.is_some() {
+            collection_authorities.unwrap()
```
StanChe, line 168: why not return it above? like 
```
let mpl_core_collections = collection_authorities.unwrap_or_else(||self.asset_collection_data....)
```

rocks-db/src/dump_client.rs
```diff
@@ -199,6 +200,25 @@ impl Storage {
             synchronizer_metrics.clone(),
         ));
 
+        let mut collections = HashMap::new();
```
StanChe, line 203: we need only the core collections, this way you may save 99.99% of the memory for this map

StanChe: so the suggested approach is 
```
let pubkeys = self.asset_static_data.iter_start().filter_map(|a|a.specification_asset_class == MplCoreCollection? Some(a.pubkey):none).collect;
let collections = self.asset_collection_data
            .batch_get(pubkeys)
```

RequescoS:
Could we also add this HashMap to our API? I think it can increase the speed of requests, especially for methods that return many assets

StanChe (approved):

RequescoS (approved):

PR #290, opened 2024-10-18
Asset data in RocksDB merged into a single column

StanChe:


StanChe:
Closing in favor of #298

PR #294, opened 2024-10-23
MTG-821 Add leaf checker

n00m4d:
# What
This PR adds new Rocks CLI tool - leaf checker. Also it adds small fix to the fork cleaner.

# Why
Was found out that fork cleaner drops data from CL items for data which was not in fork, partially it happened because of missing `if` statement which this PR adds to the fork cleaner. So goal of this new tool if to find all the assets which miss it's cl_item rows and drop sequences from that updates. In other words it will create a gap in sequences which will be found and filled with `SequenceConsistentGapfiller`.

# Important
Leaf checker will be launched with stopped ingester because it requires write rights to the Rocks.

### DIFF
--- a/rocks-db/src/bin/leaf_checker/main.rs
+++ b/rocks-db/src/bin/leaf_checker/main.rs
@@ -79,6 +79,12 @@ pub async fn main() {
 
                     let max_asset_sequence = std::cmp::max(asset_dynamic_data, asset_leaf_data);
 
+                    // if seq is 0 means asset does not exist at all
+                    // found a few such assets during testing, not sure how it happened yet
+                    if max_asset_sequence == 0 {
+                        continue;
+                    }
+
                     batch.push((cl_leaf_data.cli_tree_key, max_asset_sequence));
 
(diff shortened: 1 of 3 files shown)


rocks-db/src/bin/leaf_checker/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    println!(
+        "Found {} assets with missed CL items.",
+        assets_with_missed_cl_items
+    );
+}
+
+async fn get_asset_id(tree_id: &Pubkey, nonce: &u64) -> Pubkey {
+    Pubkey::find_program_address(
+        &["asset".as_ref(), tree_id.as_ref(), &nonce.to_le_bytes()],
+        &Pubkey::from_str("BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY").unwrap(),
```
StanChe, line 123: This address will change shortly, better to leave a todo here

StanChe (approved):

PR #295, opened 2024-10-24
[MTG-777] Switching between storage instances

kstepanovdev:
The new enum `RocksDbManager` is provided. 
Its purpose is to leave the behavior of RocksDb primary mode as it was while providing an additional feature of switching secondary instances. It means, it means, in theory only one instance will be serving requests while the second will be catching up and vice versa in a perpetual loop.

### DIFF
--- a/nft_ingester/src/bin/api/main.rs
+++ b/nft_ingester/src/bin/api/main.rs
@@ -226,7 +226,7 @@ pub async fn main() -> Result<(), IngesterError> {
 
     let cloned_rx = shutdown_rx.resubscribe();
     mutexed_tasks.lock().await.spawn(async move {
-        rocks_db_manager.catch_up(shutdown_rx).await;
+        rocks_db_manager.catch_up(cloned_rx).await;
 
         Ok(())
     });
--- a/nft_ingester/src/rocks_db.rs
+++ b/nft_ingester/src/rocks_db.rs
@@ -137,7 +137,7 @@ impl RocksDbManager {
                     % 2;
                 let free_node = &duplicate_mode.rocks_db_instances[free_node_idx];
 
-                while Arc::<rocks_db::Storage>::strong_count(&free_node) > 1 {
+                while Arc::<rocks_db::Storage>::strong_count(free_node) > 1 {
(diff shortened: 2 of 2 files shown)


nft_ingester/src/rocks_db.rs
```diff
@@ -81,3 +81,70 @@ pub async fn restore_rocksdb(config: &IngesterConfig) -> Result<(), BackupServic
...
+        const SLEEP_TIME_MS: u64 = 10;
+
+        match self {
+            RocksDbManager::Primary(_) => {}
+            RocksDbManager::Secondary(duplicate_mode) => {
+                let free_node_idx =
+                    (duplicate_mode.current_rocks_db.load(Ordering::Relaxed) + 1) % 2;
+                let free_node = &duplicate_mode.rocks_db_instance[free_node_idx];
+
+                while Arc::<rocks_db::Storage>::strong_count(&free_node) > 1 {
```
RequescoS, line 131: maybe also pass shutdown_rx here? we do not need to catch up DB when stop signal is sent to app

kstepanovdev: `catch_up` will be fired only if clone_rx (main:229) is empty. Isn't it sufficient?

nft_ingester/src/bin/api/main.rs
```diff
@@ -84,24 +86,39 @@ pub async fn main() -> Result<(), IngesterError> {
...
+        &secondary_storage_first_path,
         mutexed_tasks.clone(),
         red_metrics.clone(),
         MigrationState::Last,
     )
     .unwrap();
 
-    let rocks_storage = Arc::new(storage);
+    let secondary_storage_second_path = config
+        .rocks_db_secondary_path_container
```
RequescoS, line 104: We need to use another env. Now you are using `config.rocks_db_secondary_path_container` for both `secondary_storage_second_path` and `secondary_storage_first_path` but they must be different

nft_ingester/src/rocks_db.rs
RequescoS: yeap, the comment is about a situation you described

PR #296, opened 2024-10-28
MTG-819 Add metrics for data processing time

n00m4d:
# What
This PR adds new metrics - MessageProcessMetricsConfig.

# Why
It will allow us to see how fast indexer process data received from Geyser plugin.

# How
Metric call was added to main accounts and transactions processing functions. By main functions I mean functions where we extract data from Redis(or TCP connection), call processor and then call acknowledge. It means that during backfill processes that metrics will not be changed.

### DIFF
--- a/metrics_utils/src/lib.rs
+++ b/metrics_utils/src/lib.rs
@@ -123,8 +123,7 @@ pub struct MetricLabelWithStatus {
 
 #[derive(Debug, Clone)]
 pub struct MessageProcessMetricsConfig {
-    accounts_read: Family<MetricLabel, Histogram>,
-    transactions_read: Family<MetricLabel, Histogram>,
+    data_read: Family<MetricLabel, Histogram>,
 }
 
 impl Default for MessageProcessMetricsConfig {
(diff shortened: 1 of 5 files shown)


metrics_utils/src/lib.rs
```diff
@@ -119,6 +121,61 @@ pub struct MetricLabelWithStatus {
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl MessageProcessMetricsConfig {
+    pub fn new() -> Self {
+        Self {
+            accounts_read: Family::<MetricLabel, Histogram>::new_with_constructor(|| {
+                Histogram::new(exponential_buckets(1.0, 1.8, 10))
```
StanChe, line 140: I'd revise this, as it'll have buckets up to 357ms, which seems to be too low

metrics_utils/src/lib.rs
```diff
@@ -119,6 +121,61 @@ pub struct MetricLabelWithStatus {
     pub status: MetricStatus,
 }
 
+#[derive(Debug, Clone)]
+pub struct MessageProcessMetricsConfig {
+    accounts_read: Family<MetricLabel, Histogram>,
```
StanChe, line 126: do we need 2, or just an additional label for account/transaction maybe?

StanChe: Below you're using this label already, so this may be the same family

StanChe (commented):
Great job, thank you

PR #297, opened 2024-10-30
MTG-845 Fix for fork cleaner and boost for leaf checker

n00m4d:
# What
This PR brings 2 changes:

- fix for fork cleaner. Now it will not compete with gap filler

- optimisations for leaf checker.

# Why
## Fork cleaner
Before these changes fork cleaner could drop some data from cl items column family because it was not synced with gap filler. Here is description how could it happen.

There was a TX in the network which appeared in fork and main chain, so Geyser sends us 2 same TXs with same sequences but different slots. And first TX we've got was forked one. Fork cleaner dropped sequences related to it from tree_seq_idx at this time in parallel thread sequence consistent checker found this gap in tree_seq_idx and saves slots for parse in force_reingestable_slots. Slot_force_persister processes that slot but data were not merged to CLItems because fork cleaner not finished first iterator and there are still forked data with same seq in CLItems so merge operation simply not executed, because sequence in new value has to be bigger. Then fork cleaner finish first iterator and start to iterate over CLItems and drop forked data, but new data will not be written for that asset because slot_force_persister already done it's job in parallel.

Now fork cleaner will iterate over ClItems first and drop data from there and only then will drop data from tree_seq_idx. With this flow we are save from situation when two consistency check workers competing with each other because gapfiller will look only at gaps in sequences but not ClItems column.

## Leaf checker
There were made some small optimisation changes which speed up leaf checker on ~48%. Instead of selecting data in one thread not we process data with N tokio green threads. Also there was an attempt to use only Rocks batch selects and leave application single threaded but it did not give any good results.

### DIFF
--- a/rocks-db/src/bin/leaf_checker/main.rs
+++ b/rocks-db/src/bin/leaf_checker/main.rs
@@ -44,7 +44,10 @@ pub async fn main() {
 
     println!("Start looking for a missed CL items...");
 
-    let mut batch = Vec::new();
+    let mut delete_batch = Vec::new();
+
+    let mut select_batch_keys = Vec::new();
+    let mut select_batch_data = Vec::new();
 
     for cl_leaf in source_db.cl_leafs.iter_start() {
         match cl_leaf {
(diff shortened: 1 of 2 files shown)


rocks-db/src/bin/leaf_checker/main.rs
```diff
@@ -44,53 +44,24 @@ pub async fn main() {
...
-                    let max_asset_sequence = std::cmp::max(asset_dynamic_data, asset_leaf_data);
-
-                    // if seq is 0 means asset does not exist at all
-                    // found a few such assets during testing, not sure how it happened yet
-                    if max_asset_sequence == 0 {
-                        continue;
-                    }
+                select_batch.push(value.to_vec());
+
+                if select_batch.len() >= 2500 {
```
StanChe, line 56: This deserves to be a constant

rocks-db/src/bin/leaf_checker/main.rs
```diff
@@ -117,6 +94,88 @@ pub async fn main() {
...
+}
+
+async fn process_leaf(storage: Arc<Storage>, data: Vec<u8>) -> Option<(Pubkey, u64)> {
+    let cl_leaf_data = deserialize::<ClLeaf>(data.as_ref()).unwrap();
+
+    let key = ClItemKey::new(cl_leaf_data.cli_node_idx, cl_leaf_data.cli_tree_key);
+
+    let cl_item_rows = storage
+        .cl_items
+        .batch_get(vec![key])
```
StanChe, line 136: batch_get allows you to select a batch that'll return None if there is no value, no need to select individually. Also you're more interested in missing keys, rather then values, rocks has some methods to work with those

n00m4d: thanks, done it.
it gave additional 6% boost

StanChe (approved):

RequescoS (approved):

PR #299, opened 2024-11-01
[mtg-862] Make columns copier execute tasks asynchronously

kstepanovdev:


kstepanovdev:
Closed in prior of #300

PR #300, opened 2024-11-01
updated column copier with proper async and progress bar

StanChe:


rocks-db/src/bin/column_copier/main.rs
```diff
@@ -75,38 +76,60 @@ async fn copy_column_families(
...
                 cf_name,
                 start_column.elapsed().as_secs()
-            );
-        })
+            ));
+        }));
     });
-    for task in tasks {
-        task.await.unwrap();
+    for _ in 0..columns_to_copy.len() {
```
kstepanovdev, line 131: You already awaiting the function in the main(), it cannot be executed partially. Therefore, it's redundant.

rocks-db/src/bin/column_copier/main.rs
```diff
@@ -75,38 +76,60 @@ async fn copy_column_families(
         MigrationState::Last,
     )
     .map_err(|e| e.to_string())?;
+    let mut set = JoinSet::new();
 
-    let tasks = columns_to_copy.iter().cloned().map(|cf_name| {
+    // Create a MultiProgress to manage multiple progress bars
+    let m = Arc::new(MultiProgress::new());
```
kstepanovdev, line 82: Could you please give it a meaningful name?

rocks-db/src/bin/column_copier/main.rs
StanChe: but this function will exit immediately, right? we should join the set somehow, shouldn't we?

kstepanovdev (approved):

PR #312, opened 2024-11-13
Make big table optional

n00m4d:


nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -189,9 +190,24 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        if let Some(bg_creds) = args.big_table_credentials {
+            Arc::new(BackfillSource::Bigtable(Arc::new(
+                BigTableClient::connect_new_with(bg_creds, args.big_table_timeout)
+                    .await
+                    .expect("expected to connect to big table"),
+            )))
+        } else {
+            Arc::new(BackfillSource::Rpc(rpc_client.clone()))
+        }
+    };
```
StanChe, line 209: please move back after the ctrl+c handler, so the declaration is not far from the usage

StanChe (approved):

PR #310, opened 2024-11-12
Rework of storage [mtg-868]

StanChe:

PR #313, opened 2024-11-14
MTG-886 Change config for TCP and Redis messengers

n00m4d:
# What
This PR changes config for TCP and Redis messengers a little bit. Before these changes there were 3 different config variables for messenger, now only one.

In .env.example there is comment which describes how to write the config now.

Also main func was slightly changed. Now launching workers for accounts and transactions updates listening is more structured and defined in one place.


StanChe:
We definitely have a conflicting changes in #310

n00m4d:
Close this PR because it doesn't make any sense anymore. Before that there were two receivers: Redis and TCP. But now Aura ingester supports only Redis.

PR #316, opened 2024-11-19
MTG-703 Adding peer to peer consistency checks

snorochevskiy:
This pull request contains full implementation of peer to peer2 consistency checking and missing blocks fetching for bubblegum and account nfts.

It includes:
* storing of change records for newly received bubblegum changes and account NFT accounts
* calculation of bubblegum epochs checksums and account buckets checksums at the end of the "epoch" (10000 slots)
* periodical requesting of checksums from peers, comparison with own checksums, search for missing changes, and requesting of identified missed blocks from peers


interface/src/checksums_storage.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub checksum: Option<[u8; 32]>,
+}
+
+/// Data transfer object for account NFT last received state.
+#[derive(Debug, Clone, PartialOrd, Ord)]
+pub struct AccLastChange {
+    pub account_pubkey: Pubkey,
+    pub slot: u64,
+    pub write_version: u64,
+    pub data_hash: u64,
```
RequescoS, line 68: As far as we are tracking not only slot+write_version for accounts but also data_hash, we need to discuss the consensus mechanism. 

If i understood correctly, the current implementation would identify accounts with the same slot+write_version but different data_hashes as different updates and will try to synchronize them between nodes. However the current accounts processing mechanism will not allow updating existing accounts with the same  slot+write_version. So nodes will try to receive "updates" from each other, send these updates to each other but not process them correctly. 

Maybe this sounds like a separate task but we need to discuss approach that we think will be the better

RequescoS, line 68: Also, we do not track data_hashes for transaction-based protocol. Maybe we need also them here? I think it makes sense, because it will give us additional 'resist' from Solana forks

nft_ingester/src/message_parser.rs
```diff
@@ -193,6 +194,7 @@ impl MessageParser {
                     slot_updated: account_update.slot as i64,
                     amount: ta.amount as i64,
                     write_version: account_update.write_version,
+                    data_hash: calc_solana_account_data_hash(&account_update.data),
```
RequescoS, line 197: Could we approximately calculate the performance impact of calculating hashes "in place" for each account? We have many account updates, way larger updates amount than we have for transactions, so it would be great to understand if this increases account processing time

nft_ingester/src/bin/ingester/main.rs
```diff
@@ -694,15 +709,34 @@ pub async fn main() -> Result<(), IngesterError> {
...
+
+    consistency_calculator::run_bg_consistency_calculator(
+        nft_change_rcv,
+        primary_rocks_storage.clone(),
+        force_reingestable_slot_processor.clone(),
+        shutdown_rx.resubscribe(),
+        metrics_state.checksum_calculation_metrics.clone(),
+    );
+
+    if let Some(peer_urls_file) = config.peer_urls_file.as_ref() {
```
RequescoS, line 729: Do we need to add mounting for this file in docker compose?

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+    {
+        tokio::time::sleep(Duration::from_secs(10)).await;
+    }
+
+    let grand_epoch = grand_epoch_of_epoch(epoch);
+
+    let mut missing_bbgm_changes: HashMap<BbgmChangeRecord, HashSet<usize>> = HashMap::new();
+    let trusted_peers = peers_provider.list_trusted_peers().await;
+
+    for (peer_ind, trusted_peer) in trusted_peers.iter().enumerate() {
```
RequescoS, line 179: Maybe we need to try to make this process more async? Here many i/o with other peers, so maybe it makes sense to spawn separate tasks for communicating with each peer?

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+        .observe(start.elapsed().as_secs_f64());
+
+    let ge_cmp_res = cmp(&my_ge_chksms, &peer_ge_chksms);
+    let ge_trees_to_check = ge_cmp_res
+        .we_miss
+        .iter()
+        .chain(ge_cmp_res.different.iter())
+        .map(|&a| a.tree_pubkey)
+        .collect::<Vec<_>>();
+    for tree_pk in ge_trees_to_check {
```
RequescoS, line 264: Maybe processing each grand_epoch also deserves to be spawned in separate task because of many i/o here

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+            result.extend(changes_cmp_res.we_miss.into_iter().map(|a| a.to_owned()));
+            // TODO: track different
+        }
+    }
+
+    result
+}
+
+#[allow(clippy::while_let_on_iterator)]
+async fn handle_missing_bbgm_changes(
```
RequescoS, line 318: Point to discuss: a consensus mechanism 

If we have many hosts in the network we need to think about not just fetching any updates from each host but finding consensus between them and rejecting updates from some hosts if they contradict the majority of the network

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+                let Ok(peer_client) =
+                    grpc::client::Client::connect_to_url(&trusted_peers[*peer_ind]).await
+                else {
+                    continue;
+                };
+                clients.insert(*peer_ind, peer_client);
+                clients.get_mut(peer_ind).unwrap()
+            };
+            if let Ok(block) = client
+                .get_block(change.slot, Option::<Arc<grpc::client::Client>>::None)
```
RequescoS, line 344: We fetching the whole block, without indicating which trees we want to receive. In the future we need to add methods for syncing on concrete trees

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+        Err(e) => {
+            tracing::error!("Error reading grand epochs from DB: {}", e.to_string());
+            return Vec::new();
+        }
+    };
+    metrics
+        .db_bubblegum_get_grand_epochs_latency
+        .observe(start.elapsed().as_secs_f64());
+
+    let ge_cmp_res = cmp(&my_ge_chksms, &peer_ge_chksms);
```
RequescoS, line 257: Maybe we need to add some config for understanding which trees we are indexing. Because for now if i understood right we may mark as missed trees that we do not want to index

nft_ingester/src/consistency_calculator.rs
```diff
@@ -0,0 +1,833 @@
...
+            let metrics = metrics.clone();
+            tokio::spawn(async move {
+                process_acc_tasks(storage, acc_tasks, metrics).await;
+            })
+        };
+
+        loop {
+            let calc_msg = tokio::select! {
+                msg = rcv.recv() => msg,
+                _ = shutdown_signal.recv() => {
```
RequescoS, line 249: nit: as far as this runs inside simple tokio task that is not related to any JoinSet, shutdown_signal is not really needed here because we do not wait for this task to complete anywhere

nft_ingester/src/message_parser.rs
snorochevskiy: According to my tests, it takes ~ 9 microseconds to calculate hash for 1KB of data, which looks acceptable.
(anyway xxhash is one of most performant hashing algorithms)

interface/src/checksums_storage.rs
snorochevskiy: This is purely for account NFTs, because for account we don't have cleaning of forks, and as the result we cannot rely on slot. That's why we ignore slots and take only `pubkey+write_version_data_hash` when we calculate account changes checksum, and when we compare peer's account change with local account change.
If the `slot` is different, but the `data_hash` is same, we assume that's the same account change.

snorochevskiy: IMO, we won't gain anything by adding `data_hash` to bubblegum, since we do already have forks cleaner and sequence consistency checks for it.

nft_ingester/src/bin/ingester/main.rs
snorochevskiy: Probably, we do...

nft_ingester/src/consistency_bg_job.rs
snorochevskiy: This can be a good improvement in the future, but for the "phase 1", I'd prefer to run it sequentially, and collect CPU and disk metrics to understand whether we have some potential bandwidth for parallelization.
Cause we can easily fall into a scenario when we are stressing the disk with bunch of concurrent peer-2-peer consistency checks, and as the result the indexing part lacks resources for fulfilling it's main duty.

nft_ingester/src/consistency_bg_job.rs
snorochevskiy: And same here: a potentially good improvement in the future.

nft_ingester/src/consistency_bg_job.rs
snorochevskiy: Right, this could be a good thing for interaction with non-trusted peers, but we've decided to not make it as part of the initial implementation.

entities/src/models.rs
```diff
@@ -610,6 +619,26 @@ pub struct UnprocessedAccountMessage {
     pub id: String,
 }
 
+impl UnprocessedAccountMessage {
+    pub fn solana_change_info(&self) -> (Pubkey, u64, u64, u64) {
+        let (slot, write_version, data_hash) = match &self.account {
+            UnprocessedAccount::MetadataInfo(v) => (v.slot_updated, v.write_version, v.data_hash),
+            UnprocessedAccount::Token(v) => (v.slot_updated as u64, v.write_version, v.data_hash),
```
kstepanovdev, line 626: Are we ok with converting `i64` as `u64`? In case it cannot be negative, why then `TokenAccount` stores it as i64? Some kind of restrictions from the DB?

grpc/proto/consistency_api.proto
```diff
@@ -0,0 +1,166 @@
...
+    rpc ProposeMissingBbgmChanges(BbgmChangeList) returns (google.protobuf.Empty);
+}
+
+service AccConsistencyService {
+    rpc GetAccGrandBucketChecksums(google.protobuf.Empty) returns (AccGrandBucketChecksumsList);
+    rpc GetAccBucketChecksumsInGrandBucket(GetAccBucketsReq) returns (AccBucketChecksumsList);
+    rpc GetAccsInBucket(GetAccReq) returns (AccList);
+
+    rpc ProposeMissingAccChanges(AccList) returns (google.protobuf.Empty);
+}
```
kstepanovdev, line 166: nit: newline

grpc/src/client.rs
```diff
@@ -21,8 +21,11 @@ pub struct Client {
 
 impl Client {
     pub async fn connect(peer_discovery: impl PeerDiscovery) -> Result<Self, GrpcError> {
-        let url = Uri::from_str(peer_discovery.get_gapfiller_peer_addr().as_str())
-            .map_err(|e| GrpcError::UriCreate(e.to_string()))?;
+        Client::connect_to_url(peer_discovery.get_gapfiller_peer_addr().as_str()).await
```
kstepanovdev, line 24:
```suggestion
Client::connect_to_url(&peer_discovery.get_gapfiller_peer_addr()).await
```

just a preference of style, feel free to ignore

kstepanovdev, line 24:
```suggestion
        Client::connect_to_url(&peer_discovery.get_gapfiller_peer_addr()).await
```

Just a preference of style, feel free to ignore

interface/src/checksums_storage.rs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        for piece in data {
+            piece.hash(state)
+        }
+    }
+}
+
+/// Interface for querying bubblegum checksums from peer
+/// or local storage.
+#[async_trait]
```
kstepanovdev, line 100: Isn't `async trait` stabilized?

metrics_utils/src/lib.rs
```diff
@@ -230,6 +242,319 @@ impl BackfillerMetricsConfig {
...
+    pub db_account_get_latests_latency: Histogram,
+
+    pub bbgm_consistency_grpc_client: Arc<BubblegumConsistencyGrpcClientMetricsConfig>,
+    pub acc_consistency_grpc_client: Arc<AccoountConsistencyGrpcClientMetricsConfig>,
+
+    pub found_missing_bubblegums: Gauge,
+    pub found_missing_accounts: Gauge,
+}
+
+impl Default for Peer2PeerConsistencyMetricsConfig {
```
kstepanovdev, line 261: It seems to me such default has no usage since `new()` can be freely called instead. Seems like only one should survive imo.

From my perspective, new() with no parameters = default (if the function doesn't provoke side effects)

metrics_utils/src/lib.rs
```diff
@@ -230,6 +242,319 @@ impl BackfillerMetricsConfig {
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Peer2PeerConsistencyMetricsConfig {
+    pub fn new() -> Peer2PeerConsistencyMetricsConfig {
+        let mk_histogram = || Histogram::new(exponential_buckets(20.0, 1.8, 10));
```
kstepanovdev, line 269: Those numbers are slightly magical. What do they mean?

interface/src/checksums_storage.rs
snorochevskiy: rust-toolchain.toml:
```toml
[toolchain]
channel = "1.75.0"
```
🙃

interface/src/checksums_storage.rs
```diff
@@ -0,0 +1,149 @@
+use async_trait::async_trait;
+use solana_sdk::pubkey::Pubkey;
+
+/// Type of checksum for bubblegum epochs and account NFT buckets.
+/// It is technically a SHA3 hash.
+pub type Chksm = [u8; 32];
```
StanChe, line 6: I thought the type above was for this
```suggestion
    pub checksum: Option<Chksm>,
```

interface/src/checksums_storage.rs
StanChe: For the consensus part - that's for us to define. Essentially, we should expect the following:
- same write_version, same slot, same hash - all good
- same write_version, same slot, different hash - we need the consensus to define, what the correct version is. I'd suggest we ask RPC of the current state (if it has not changed yet, of course) 
- same write_version, different slot, same hash - all good, take the highest slot on merge probably
- same write_version, different slot, different hashes - looks like a fork, take the highest slot on merge
- different write_version, same/different slot, same/different hash - take the highest write_version

metrics_utils/src/lib.rs
```diff
@@ -230,6 +242,319 @@ impl BackfillerMetricsConfig {
...
 }
 
+#[derive(Debug, Clone)]
+pub struct Peer2PeerConsistencyMetricsConfig {
+    pub db_bubblegum_get_grand_epochs_latency: Histogram,
+    pub db_bubblegum_get_epochs_latency: Histogram,
+    pub db_bubblegum_get_changes_latency: Histogram,
+    pub db_account_get_grand_buckets_latency: Histogram,
+    pub db_account_get_buckets_latency: Histogram,
+    pub db_account_get_latests_latency: Histogram,
```
StanChe, line 252: Why not a single Family<MetricLabel, Histogram>?

metrics_utils/src/lib.rs
```diff
@@ -230,6 +242,319 @@ impl BackfillerMetricsConfig {
...
+pub struct BubblegumConsistencyGrpcClientMetricsConfig {
+    pub peers_bubblegum_get_grand_epochs_for_tree_latency: Histogram,
+    pub peers_bubblegum_get_grand_epochs_latency: Histogram,
+    pub peers_bubblegum_get_epochs_latency: Histogram,
+    pub peers_bubblegum_get_changes_latency: Histogram,
+
+    pub peers_bubblegum_get_grand_epochs_for_tree_errors: Family<MetricLabel, Counter>,
+    pub peers_bubblegum_get_grand_epochs_errors: Family<MetricLabel, Counter>,
+    pub peers_bubblegum_get_epochs_errors: Family<MetricLabel, Counter>,
+    pub peers_bubblegum_get_changes_errors: Family<MetricLabel, Counter>,
```
StanChe, line 345: it looks more like an added label to me. Peers_sync_latency(protocol: "bubblegum/account", method/endpoint: "get_grand_epochs/get_epochs/get_changes"

nft_ingester/tests/consistency_bg_job_test.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[tokio::test]
+    pub async fn test_checksum_exchange() {
+        let my_storage = RocksTestEnvironment::new(&[]).storage;
+
+        let peer_storage = RocksTestEnvironment::new(&[]).storage;
+
+        // prepare
+        let tree1 = Pubkey::new_unique();
+
+        // This change is for epoch we won't calculate in the test,
```
StanChe, line 20: the comment is not valid in this context

nft_ingester/tests/consistency_bg_job_test.rs
```diff
@@ -0,0 +1,99 @@
...
+        let k1_2 = BubblegumChangeKey::new(tree1, 10112, 3);
+        let v1_2 = BubblegumChange {
+            signature: "3".to_string(),
+        };
+        peer_storage
+            .bubblegum_changes
+            .put(k1_2.clone(), v1_2.clone())
+            .unwrap();
+
+        // This will be also ignored
```
StanChe, line 50: this one as well

nft_ingester/tests/consistency_calculator_test.rs
```diff
@@ -0,0 +1,255 @@
...
+        );
+    }
+
+    #[tokio::test]
+    async fn test_calc_acc_nft_checksums() {
+        let storage = RocksTestEnvironment::new(&[]).storage;
+
+        let acc_change_val = AccountNftChange {};
+
+        let acc1_pubkey = make_pubkey_in_bucket(0);
```
StanChe, line 115: why not just unique? is there any logic behind those keys?

StanChe: bucket, got it

metrics_utils/src/lib.rs
snorochevskiy: I usually prefer to have a set of plain metrics, instead of one labeled, because it is much easier to query them by other monitoring systems. But sure, I can turn these into a family. Should I?

interface/src/checksums_storage.rs
snorochevskiy:
```rust
pub type Chksm = [u8; 32];
```
is just an alias for a potential change of checksum type in the future (that will never happen)

snorochevskiy: I believe it will just introduce a hell of wrap and unroll calls 🥲

kstepanovdev: Deref for the rescue?

kstepanovdev: Just thinking aloud tho, not a call to action

interface/src/checksums_storage.rs
snorochevskiy: Yes, but `#[async_trait]` is still required to build trait object

StanChe (commented):
Great work, thank you. Several open questions need clarification and the metrics should be simplified/moved to more appropriate measuring places.

nft_ingester/tests/consistency_calculator_test.rs
```diff
@@ -0,0 +1,255 @@
...
+            .put(acc2_change_key1.clone(), acc_change_val.clone())
+            .unwrap();
+
+        // SUT
+        calc_acc_nft_checksums(&storage, 0).await;
+
+        // Verify account last state updated
+        let latest_acc1_key = AccountNftKey::new(acc1_pubkey);
+        let latest_acc1_val = storage
+            .acc_nft_last
```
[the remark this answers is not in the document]
snorochevskiy: It is the last seen change of the account with the given pubkey.

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+
+impl FileSrcAuraPeersProvides {
+    pub fn new(file_path: String) -> FileSrcAuraPeersProvides {
+        FileSrcAuraPeersProvides { file_path }
+    }
+}
+
+#[async_trait::async_trait]
+impl AuraPeersProvides for FileSrcAuraPeersProvides {
+    async fn list_trusted_peers(&self) -> Vec<String> {
```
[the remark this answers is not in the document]
snorochevskiy: That's to make possible to change the list of peers without restarting the appliction

rocks-db/src/storage_consistency.rs
```diff
@@ -0,0 +1,909 @@
...
+            }
+        }
+        Ok(result)
+    }
+
+    async fn propose_missing_changes(
+        &self,
+        _changes: Vec<interface::checksums_storage::AccLastChange>,
+    ) {
+        // Do nothing
```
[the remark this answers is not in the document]
snorochevskiy: A little bit confusing text copy... we may do [MTG-1039]

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+async fn exchange_bbgms_with_peers(
+    epoch: u32,
+    storage: Arc<Storage>,
+    peers_provider: Arc<dyn AuraPeersProvides + Send + Sync>,
+    force_reingestable_slot_processor: Arc<dyn BlockConsumer + Send + Sync>,
+    metrics: Arc<Peer2PeerConsistencyMetricsConfig>,
+) {
+    tracing::info!("Starting bubblegum changes peer-to-peer exchange for epoch={epoch}");
+    while get_calculating_bbgm_epoch()
+        .map(|e| e == epoch)
```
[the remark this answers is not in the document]
snorochevskiy: Theoretically that should not happen, but I've changed it to compare current epoch with the last calculated.

nft_ingester/src/consistency_bg_job.rs
```diff
@@ -0,0 +1,743 @@
...
+
+            let start = Instant::now();
+            let my_changes = match we.list_epoch_changes(epoch, tree_pubkey, None, None).await {
+                Ok(v) => v,
+                Err(e) => {
+                    tracing::error!("Error reading bubblegum changes from DB: {}", e.to_string());
+                    return result;
+                }
+            };
+            metrics
```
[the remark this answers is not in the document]
snorochevskiy: This one measures how much time it takes on our side to fetch the data.
On the GRPC client side (BbgmConsistencyApiClientImpl and AccConsistencyApiClientImpl) there are separate metrics that measure how much time it takes to call the peer.

rocks-db/src/storage_consistency.rs
```diff
@@ -0,0 +1,909 @@
...
+            if result.len() >= max_result {
+                break;
+            }
+        }
+
+        Ok(result)
+    }
+
+    async fn propose_missing_changes(&self, _changes: &[BbgmChangeRecord]) {
+        // TODO: how handle?
```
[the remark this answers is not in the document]
StanChe: https://linear.app/mplx/issue/MTG-1039/peer2peer-consistency-propose-changes-to-peer

PR #322, opened 2024-11-25
[MTG-964] Ownership model fix using a history of owners

StanChe:
adding a history of owners for an asset, tracking any ownership change to determine the correct owner

As a next steps we could consider storing any owner change, so cNFTs will also have the history available.


grpc/src/mapper.rs
```diff
@@ -191,6 +191,8 @@ impl TryFrom<AssetDetails> for CompleteAssetDetails {
                 .transpose()?
                 .ok_or(GrpcError::MissingField("owner_type".to_string()))?,
             owner_delegate_seq,
+            is_current_owner: todo!(),
```
kstepanovdev, line 194: I suppose it will panic on `try_into()` conversion

StanChe: Yep, grpc will panic: https://linear.app/mplx/issue/MTG-977/propagate-ownership-changes-to-grpc made a ticket for that

rocks-db/src/flatbuf/asset.fbs
```diff
@@ -175,6 +172,7 @@ table AssetCompleteDetails {
   authority: AssetAuthority;
   owner: AssetOwner;
   collection: AssetCollection;
+  other_known_owners: [AssetOwner];
 }
 
 root_type AssetCompleteDetails;
```
kstepanovdev, line 178: ultra nit: newline

kstepanovdev (approved):
Found no problems in syntax, lgtm

PR #324, opened 2024-11-25
MTG-959 fix a few issues with core assets

n00m4d:
# What

This PR fixes a few issues with Core assets.

1. If there is no authority for asset it will not return system prog adress but empty string instead.

2. For Core collections API doesn't check collection verification field.

3. If there is no collection for asset `showUnverifiedCollection` flag should not affect the API response.

4. Please take a closer look at changes in `nft_ingester/src/processors/account_based/mpl_core_processor.rs`. It was made to fix plugin ordering for assets. Kinda minor stuff so maybe we don't need it because it adds convertion from HashMap to BTreeMap during asset processing time but without that change same plugins list can be shown in different order on API side.

### DIFF
--- a/nft_ingester/src/api/dapi/asset.rs
+++ b/nft_ingester/src/api/dapi/asset.rs
@@ -3,6 +3,7 @@ use std::string::ToString;
 use std::sync::Arc;
 
 use entities::api_req_params::{AssetSortDirection, Options};
+use entities::enums::SpecificationAssetClass;
 use entities::models::{AssetSignatureWithPagination, OffChainData};
 use interface::asset_sigratures::AssetSignaturesGetter;
 use interface::json::{JsonDownloader, JsonPersister};
@@ -73,7 +74,7 @@ fn convert_rocks_asset_model(
         asset_leaf: leaf,
         offchain_data,
         asset_collections: data.collection.clone(),
-        assets_authority: data.authority.clone().unwrap_or(AssetAuthority::default()),
+        assets_authority: data.authority.clone(),
         edition_data: static_data
             .edition_address
             .and_then(|e| asset_selected_maps.editions.get(&e).cloned()),
@@ -135,17 +136,18 @@ fn asset_selected_maps_into_full_asset(
     options: &Options,
 ) -> Option<FullAsset> {
     if !options.show_unverified_collections {
-        if let Some(collection_data) = asset_selected_maps
-            .asset_complete_details
-            .get(id)
-            .and_then(|a| a.collection.clone())
-        {
-            if !collection_data.is_collection_verified.value {
-                return None;
+        if let Some(asset_complete_details) = asset_selected_maps.asset_complete_details.get(id) {
+            if let Some(asset_static_details) = &asset_complete_details.static_details {
+                // collection itself cannot have a collection
+                // TODO!: should we also include in this check FungibleToken?
+                if &asset_static_details.specification_asset_class != &SpecificationAssetClass::MplCoreCollection {
+                    if let Some(collection_details) = &asset_complete_details.collection {
+                        if !collection_details.is_collection_verified.value {
+                            return None;
+                        }
+                    }
+                }
             }
-        } else {
-            // don't have collection data == collection unverified
-            return None;
         }
     }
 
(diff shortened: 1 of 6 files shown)


nft_ingester/src/processors/account_based/mpl_core_processor.rs
```diff
@@ -76,9 +77,9 @@ impl MplCoreProcessor {
         // If it is an `Address` type, use the value directly.  If it is a `Collection`, search for and
         // use the collection's authority.
         let update_authority = match asset.update_authority {
-            UpdateAuthority::Address(address) => address,
+            UpdateAuthority::Address(address) => Some(address),
             UpdateAuthority::Collection(address) => storage.get_authority(address),
-            UpdateAuthority::None => Pubkey::default(),
+            UpdateAuthority::None => Some(Pubkey::default()),
```
StanChe, line 82: Why are we using default here instead of None?

nft_ingester/src/processors/account_based/mpl_core_processor.rs
```diff
@@ -133,7 +134,14 @@ impl MplCoreProcessor {
             })
             .unwrap_or((0, &default_creators));
 
-        let mut plugins_json = serde_json::to_value(&asset.plugins)
+        // convert HashMap plugins into BTreeMap to have always same plugins order
+        // for example without ordering 2 assets with same plugins can have different order saved in DB
+        // it affects only API response and tests
+        let ordered_plugins: BTreeMap<_, _> = asset.plugins
```
StanChe, line 140: wdyt of sorting when you're building the response? The json and order are a bit tricky, as you're implicitly relying on the reading code not to decode it as a map again.

nft_ingester/src/processors/account_based/mpl_core_processor.rs
```diff
@@ -254,12 +262,16 @@ impl MplCoreProcessor {
             created_at: account_data.slot_updated as i64,
             edition_address: None,
         });
-        models.asset_authority = Some(AssetAuthority {
-            pubkey: asset_key,
-            authority: update_authority,
-            slot_updated: account_data.slot_updated,
-            write_version: Some(account_data.write_version),
-        });
+        if let Some(upd_auth) = update_authority {
```
StanChe, line 265: this one may be None only for collections, and is not none if the asset.update_authority is none. Is this how it's intended to work?

nft_ingester/src/processors/account_based/mpl_core_processor.rs
```diff
@@ -133,7 +134,14 @@ impl MplCoreProcessor {
             })
             .unwrap_or((0, &default_creators));
 
-        let mut plugins_json = serde_json::to_value(&asset.plugins)
+        // convert HashMap plugins into BTreeMap to have always same plugins order
+        // for example without ordering 2 assets with same plugins can have different order saved in DB
+        // it affects only API response and tests
+        let ordered_plugins: BTreeMap<_, _> = asset.plugins
+            .iter()
+            .map(|(key, value)| (format!("{:?}", key), value))
```
StanChe, line 142: why format here? doesn't serve properly encode the PluginType? Looks like `asset.plugins.iter().collect()` should have worked, didn't it?

n00m4d: no it's not. Because HashMap as a key has `PluginType` type and that type doesn't implement `Ord`. But this convertion requires this type to be `Ord`

StanChe (approved):

nft_ingester/src/api/dapi/asset.rs
```diff
@@ -135,17 +136,18 @@ fn asset_selected_maps_into_full_asset(
...
-            .get(id)
-            .and_then(|a| a.collection.clone())
-        {
-            if !collection_data.is_collection_verified.value {
-                return None;
+        if let Some(asset_complete_details) = asset_selected_maps.asset_complete_details.get(id) {
+            if let Some(asset_static_details) = &asset_complete_details.static_details {
+                // collection itself cannot have a collection
+                // TODO!: should we also include in this check FungibleToken?
+                if &asset_static_details.specification_asset_class != &SpecificationAssetClass::MplCoreCollection {
```
RequescoS, line 143: Can other protocols have collections as separate assets/accounts? If so, maybe we can check it somehow (i mean check if the account/asset represents a collection). Maybe it sounds like a separate ticket, so there are no actions required now

RequescoS (approved):

PR #325, opened 2024-11-26
[mtg-778] Clean asset slot update idx

kstepanovdev:
After each synchronization redundant idxs will be removed from the rocksDb

### DIFF
--- a/nft_ingester/src/bin/ingester/main.rs
+++ b/nft_ingester/src/bin/ingester/main.rs
@@ -815,7 +815,6 @@ pub async fn main() -> Result<(), IngesterError> {
                             }
                             Err(e) => {
                                 error!("Failed to clean synchronized indexes for {:?} with error {}", asset_type, e);
-                                break;
                             }
                         }
                         tokio::time::sleep(Duration::from_secs(SECONDS_TO_RETRY_IDXS_CLEANUP)).await;


rocks-db/src/asset_client.rs
```diff
@@ -113,6 +113,45 @@ impl Storage {
...
+                let from = key_type_pairs[0].0.clone();
+                (from, cf)
+            }
+        };
+
+        if from == last_synced_key {
+            return Ok(());
+        }
+
+        let mut batch = rocksdb::WriteBatchWithTransaction::<false>::default();
```
RequescoS, line 127: Can you please explain the reason for using batch here? It seems like you perform a single operation on this batch, so does it make sense to use it at all?

rocks-db/src/batch_savers.rs
```diff
@@ -270,4 +270,13 @@ impl BatchSaveStorage {
         }
         Ok(())
     }
+
+    pub fn clean_syncronized_idxs(
```
RequescoS, line 274: Do we need such wrapper functions?

nft_ingester/tests/api_tests.rs
```diff
@@ -3714,4 +3714,118 @@ mod tests {
...
+
+        while let Some(_) = idx_fungible_asset_iter.next() {
+            number_of_fungible_idxs += 1;
+        }
+        assert_eq!(number_of_fungible_idxs, 1);
+
+        while let Some(_) = idx_non_fungible_asset_iter.next() {
+            number_of_non_fungible_idxs += 1;
+        }
+        assert_eq!(number_of_non_fungible_idxs, 3);
```
RequescoS, line 3798: nit: you can use iter_start().count()

RequescoS:
Great job, thank you! Left only a couple of minor comments

rocks-db/src/asset_client.rs
kstepanovdev: Not needed anymore, indeed. Previously it was multiple deletions via iterator

nft_ingester/src/index_syncronizer.rs
```diff
@@ -172,44 +172,71 @@ where
...
             SyncStatus::RegularSyncRequired(state) => {
                 self.regular_nft_syncronize(rx, state.last_indexed_key, state.last_known_key)
-                    .await
+                    .await?;
             }
-            SyncStatus::NoSyncRequired => Ok(()),
+            SyncStatus::NoSyncRequired => {}
+        }
+
+        if let Some(encoded_key) = self.index_storage.fetch_last_synced_id(asset_type).await? {
```
RequescoS, line 193: We always expert fetch_last_synced_id() return here Some(_) because if we receive None it will mean that some jobs are incomplete or we have some bug in code. So, maybe use ok_or()? Or at least add an error log if fetch_last_synced_id returned None

RequescoS (approved):

nft_ingester/tests/api_tests.rs
```diff
@@ -3714,4 +3716,109 @@ mod tests {
...
+        let idx_fungible_asset_iter = env
+            .rocks_env
+            .storage
+            .fungible_assets_update_idx
+            .iter_start();
+        let idx_non_fungible_asset_iter = env.rocks_env.storage.assets_update_idx.iter_start();
+        assert_eq!(idx_fungible_asset_iter.count(), 1);
+        assert_eq!(idx_non_fungible_asset_iter.count(), cnt + 2);
+
+        for asset_type in ASSET_TYPES {
```
n00m4d, line 3792: NIT: wdyt about wrapping part of this stuff with function? then we should not copy past it from `ingester/main.rs`

n00m4d (approved):

nft_ingester/src/bin/ingester/main.rs
```diff
@@ -75,6 +76,7 @@ pub const DEFAULT_ROCKSDB_PATH: &str = "./my_rocksdb";
 pub const ARWEAVE_WALLET_PATH: &str = "./arweave_wallet.json";
 pub const DEFAULT_MIN_POSTGRES_CONNECTIONS: u32 = 100;
 pub const DEFAULT_MAX_POSTGRES_CONNECTIONS: u32 = 100;
+pub const SECONDS_TO_RETRY_IDXS_CLEANUP: u64 = 15;
```
StanChe, line 79: make if less often, like once in 15 mins

nft_ingester/src/bin/ingester/main.rs
```diff
@@ -800,6 +800,31 @@ pub async fn main() -> Result<(), IngesterError> {
...
+
+                match clean_syncronized_idxs(primary_rocks_storage.clone(), asset_type) {
+                    Ok(_) => {
+                        info!("Cleaned synchronized indexes for {:?}", asset_type);
+                    }
+                    Err(e) => {
+                        error!("Failed to clean synchronized indexes for {:?} with error {}", asset_type, e);
+                    }
+                }
+                tokio::time::sleep(Duration::from_secs(SECONDS_TO_RETRY_IDXS_CLEANUP)).await;
```
StanChe, line 821: how does it work with termination (ctrl+c) of the app? we usually switch on it and the shutdown_rx

rocks-db/src/asset_client.rs
```diff
@@ -113,6 +113,22 @@ impl Storage {
...
+        asset_type: AssetType,
+        last_synced_key: Vec<u8>,
+    ) -> Result<()> {
+        let cf = match asset_type {
+            AssetType::Fungible => self.fungible_assets_update_idx.handle(),
+            AssetType::NonFungible => self.assets_update_idx.handle(),
+        };
+
+        let from = vec![];
+        self.db.delete_range_cf(&cf, from, last_synced_key)?;
```
RequescoS, line 128: Maybe add some additional metrics to track the execution time of this operation?

RequescoS (approved):

StanChe (approved):
just drop that break and we're good to go

PR #326, opened 2024-11-26
MTG-941 add integration e to e tests

n00m4d:
# What

This PR adds end to end integration tests for the Aura node. Test were copied from [here](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure) and refactored some methods so it's compatible with this refactored DAS node.

Please take a closer look at changes made in nft_ingester and postgre-client because there was changed visibility level for a few functions.


postgre-client/src/lib.rs
```diff
@@ -187,4 +188,31 @@ impl PgClient {
...
+        ] {
+            self.truncate_table(&mut transaction, table).await?;
+        }
+
+        transaction.execute(sqlx::query("update last_synced_key set last_synced_asset_update_key = null where id = 1;")).await?;
+
+        self.recreate_indexes(&mut transaction).await?;
+        self.recreate_constraints(&mut transaction).await?;
+
+        transaction.commit().await.map_err(|e| e)?;
```
StanChe, line 214: fyi, those await? above will not always rollback the tx, a minor concern if its only in tests

integration_tests/README.md
StanChe, line 1: Please add from where this crate was taken to this file so we respect the authors of the original tests.

postgre-client/src/lib.rs
```diff
@@ -187,4 +188,31 @@ impl PgClient {
             .observe_request(SQL_COMPONENT, TRANSACTION_ACTION, "rollback", start_time);
         Ok(())
     }
+
+    // at the moment used in integration tests only
```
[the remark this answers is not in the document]
StanChe: or any other limitation of the build scope, will it work this way?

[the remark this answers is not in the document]
n00m4d: reused integration_tests feature, thanks

postgre-client/src/lib.rs
n00m4d: added comment, thanks for info

integration_tests/README.md
n00m4d: added, thanks

integration_tests/README.md
```diff
@@ -0,0 +1,38 @@
+# Integration Tests
+
+> [!NOTE]
+> These tests were taken from the original DAS API implementation, you can find original code in [this repository](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure).
```
StanChe, line 4:
```suggestion
> These tests were taken from the reference DAS API implementation, you can find original code in [this repository](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure).
```

StanChe (approved):

Cargo.toml
```diff
@@ -44,6 +48,7 @@ uuid = { version = "1.6.1", features = ["v4"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.16", features = ["json", "env-filter", "ansi"] }
 env_logger = "0.10.1"
+log = "0.4.17"
```
RequescoS, line 51: Previously we was remover `log` crate in favor of `tracing`. Can use `tracing` in tests or `log` is required for them?

integration_tests/README.md
```diff
@@ -0,0 +1,38 @@
...
+Also gain access to mainnet RPCs and devnet RPCs and export the URLs as follows. Currently,
+we only use these URLs for reading data and storing it locally. 
+
+```
+export DEVNET_RPC_URL=...
+export MAINNET_RPC_URL=...
+```
+
+Afterwards, you can simply run the following command to run tests:
+```cargo test```
```
RequescoS, line 22: Don't we need to add `integration_tests` feature here?

integration_tests/src/account_update_tests.rs
```diff
@@ -0,0 +1,263 @@
...
+    let mint = Pubkey::try_from("843gdpsTE4DoJz3ZoBsEjAqT8UgAcyF5YojygGgGZE1f").unwrap();
+
+    let nft_accounts = get_nft_accounts(&setup, mint).await;
+
+    let request = GetAsset {
+        id: mint.to_string(),
+        options: Options::default(),
+    };
+
+    let random_pub_key = Pubkey::try_from("1111111QLbz7JHiBTspS962RLKV8GndWFwiEaqKM").unwrap();
```
RequescoS, line 134: Can't we use `Pubkey::new_unique()` for random key generation?

integration_tests/README.md
RequescoS: Oh, sorry, i thought new tests were hidden under this flag. You are right, this change is not needed

RequescoS:
Great work, thank you!

RequescoS (approved):

Cargo.toml
n00m4d: did it, thanks

PR #327, opened 2024-11-28
MTG-980 Fix a few issues with cNFT assets

n00m4d:
# What

This PR fixes a few small issues related to cNFTs.

Also it fixes one corner case related to `was_decompressed` field. From now on that field will be setted only during cNFTs processing and it's optional.

To check if it works just run the integration tests.


grpc/src/mapper.rs
```diff
@@ -47,7 +47,7 @@ impl From<CompleteAssetDetails> for AssetDetails {
             supply: value.supply.map(|v| v.into()),
             seq: value.seq.map(|v| v.into()),
             is_burnt: Some(value.is_burnt.into()),
-            was_decompressed: Some(value.was_decompressed.into()),
+            was_decompressed: value.was_decompressed.map(|v| v.into()),
```
kstepanovdev, line 50: `.map(Into::into)`, should do the trick

nft_ingester/src/api/dapi/rpc_asset_convertors.rs
```diff
@@ -547,6 +524,63 @@ pub fn asset_to_rpc(
...
+        if was_decompressed.value {
+            return Compression {
+                eligible: false,
+                compressed: false,
+                data_hash: "".to_string(),
+                creator_hash: "".to_string(),
+                asset_hash: "".to_string(),
+                tree: "".to_string(),
+                seq: 0,
+                leaf_id: 0,
```
kstepanovdev, line 544: Looks like `Default`

nft_ingester/src/api/dapi/rpc_asset_convertors.rs
```diff
@@ -547,6 +524,63 @@ pub fn asset_to_rpc(
...
+            };
+        }
+    }
+
+    Compression {
+        eligible: full_asset.asset_dynamic.is_compressible.value,
+        compressed: full_asset.asset_dynamic.is_compressed.value,
+        leaf_id: full_asset.asset_leaf.nonce.unwrap_or(0) as i64,
```
kstepanovdev, line 552: This is a conversion with loss. Is it intended?

rocks-db/src/asset.rs
```diff
@@ -3103,7 +3103,7 @@ impl AssetDynamicDetails {
             self.supply.clone().map_or(0, |supply| supply.slot_updated),
             self.seq.clone().map_or(0, |seq| seq.slot_updated),
             self.is_burnt.slot_updated,
-            self.was_decompressed.slot_updated,
+            self.was_decompressed.clone().map_or(0, |was_decompressed| was_decompressed.slot_updated),
```
kstepanovdev, line 3106: `as_ref()` instead of `clone()` looks viable in each place in the method

rocks-db/src/migrations/spl2022.rs
```diff
@@ -52,7 +52,7 @@ pub struct AssetDynamicDetailsWithoutExtentions {
     pub supply: Option<Updated<u64>>,
     pub seq: Option<Updated<u64>>,
     pub is_burnt: Updated<bool>,
-    pub was_decompressed: Updated<bool>,
```
StanChe, line 55: this datatype (AssetDynamicDetailsWithoutExtentions) may be safely removed by the way as all the migrations are turned off at the moment

StanChe (approved):

kstepanovdev (approved):

PR #330, opened 2024-11-29
MTG-1002 Drop unused data struct

n00m4d:
# What

This PR drops AssetDynamicDetailsWithoutExtentions data struct.


kstepanovdev (approved):

rocks-db/src/lib.rs
```diff
@@ -549,14 +546,9 @@ impl Storage {
             asset::AssetDynamicDetails::NAME => {
                 let mf = match migration_state {
                     MigrationState::Version(version) => match *version {
-                        CollectionAuthorityMigration::VERSION
-                            ..=ExternalPluginsMigration::VERSION => {
+                        CollectionAuthorityMigration::VERSION => {
```
RequescoS, line 549: i guess we do not need external_plugins and collection_authority migrations too, as far as we reingest/backfill all data in CompleteAssetDetails and drop all old data here

RequescoS (approved):

rocks-db/src/lib.rs
n00m4d: If so we can create separate ticket to drop that stuff as well

PR #332, opened 2024-12-02
MTG-993 Fix one integration test

n00m4d:
# What

This PR fixes `test_account_updates` integration test.

# How

It was fixed by adding `clean_db` method and deleting `chain_mutability` value update during Mint account processing. It was always set it to `Mutable` and it was kinda mistake because it affected NFTs where mutability property if defined at Metadata account, not Mint.

All other tests were not affected by these changes.

### DIFF
--- a/integration_tests/Cargo.toml
+++ b/integration_tests/Cargo.toml
@@ -33,6 +33,6 @@ tempfile = {workspace = true}
 nft_ingester = { path = "../nft_ingester" }
 postgre-client = { path = "../postgre-client", features = ["integration_tests"] }
 metrics-utils = { path = "../metrics_utils" }
-rocks-db = { path = "../rocks-db" }
+rocks-db = { path = "../rocks-db", features = ["integration_tests"] }
 usecase = { path = "../usecase" }
 entities = { path = "../entities" }
--- a/nft_ingester/src/processors/account_based/token_updates_processor.rs
+++ b/nft_ingester/src/processors/account_based/token_updates_processor.rs
@@ -170,11 +170,6 @@ impl TokenAccountsProcessor {
                     )
                 })
                 .unwrap_or_default(),
-            chain_mutability: Some(Updated::new(
-                mint.slot_updated as u64,
-                Some(UpdateVersion::WriteVersion(mint.write_version)),
-                entities::enums::ChainMutability::Mutable,
-            )),
             onchain_data: metadata_json.map(|metadata_json| {
                 Updated::new(
                     mint.slot_updated as u64,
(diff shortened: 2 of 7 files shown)


rocks-db/src/lib.rs
```diff
@@ -789,4 +789,52 @@ impl Storage {
         }
         cf_options
     }
+
+    #[cfg(feature = "integration_tests")]
+    pub async fn clean_db(&self) {
+        let column_families_to_remove = [
```
RequescoS, line 795: maybe we can somehow reuse ‎Storage::create_cf_descriptors() function here? In order not to forget to update code in two places instead of one if we add new columns. It's just a guess, it may not be possible to reuse this function such way

RequescoS (approved):

integration_tests/Cargo.toml
```diff
@@ -33,6 +33,6 @@ tempfile = {workspace = true}
 nft_ingester = { path = "../nft_ingester" }
 postgre-client = { path = "../postgre-client", features = ["integration_tests"] }
 metrics-utils = { path = "../metrics_utils" }
-rocks-db = { path = "../rocks-db" }
+rocks-db = { path = "../rocks-db", features = ["integration_tests"] }
```
kstepanovdev, line 36: At a first glance this import looks slightly bizarre to me. Do it mean we import source with the tests attached? If so, maybe we can separate common structures into a dedicated module or something?

n00m4d: there is defined method `clean_db` with integration_tests feature, not to build it for usual usage

nft_ingester/src/processors/account_based/token_updates_processor.rs
```diff
@@ -170,11 +170,6 @@ impl TokenAccountsProcessor {
                     )
                 })
                 .unwrap_or_default(),
-            chain_mutability: Some(Updated::new(
```
StanChe, line 173: Is the mint account ever mutable or immutable? when is it defined/checked?

rocks-db/src/lib.rs
```diff
@@ -789,4 +789,52 @@ impl Storage {
         }
         cf_options
     }
+
+    #[cfg(feature = "integration_tests")]
+    pub async fn clean_db(&self) {
```
StanChe, line 794: why don't you just drop the column family and recreate it?

nft_ingester/src/processors/account_based/token_updates_processor.rs
StanChe: I'm not following

StanChe (approved):
Looks ok, as a separate effort @n00m4d will look through all the other fields that may be set based on different accounts.

PR #339, opened 2024-12-16
[mtg-1065] Keep metadata inside of mutable storages up to date with occuring changes

kstepanovdev:
The issue is metadata as a part of immutable storages may be updated freely off-chain. That leads to an issue when the indexer doesn't provide up-to-date information regarding such changes.

To rectify this, the cache of the metadata has been added along with a new type of metadata that reflects its mutability. The gist of this field is to keep track of the latest changes occurring inside of metadata. For brevity, the pseudocode is simplified: 
If `cache period + last_metadata_read > curr_time && is_mutable { update_metadata }`
Such changes do not provide an instant update, however, it may be configured.

Also, the `OffChainData` has been moved to `flatbuffers` encoding, which gives us an opportunity for further changes without writing yet another migration for RocksDb.


rocks-db/src/columns/offchain_data.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn is_mutable(&self) -> bool {
+        match self {
+            StorageMutability::Immutable => false,
+            StorageMutability::Mutable => true,
+        }
+    }
+}
+
+impl From<&str> for StorageMutability {
+    fn from(storage_mutability: &str) -> Self {
```
StanChe, line 26: What is storage_mutability in this case? Is it the uri? with ipfs it has the ipfs as protocol at times, like ipfs://<hash>, while arweave is working over https and may have www prefix as well
Some examples:
ipfs://bafybeib3ogmisx45y45er4jjxi3nd7cta5smeapdoueodkt363g4loeisa/36515.json
"https://www.arweave.net/Hkldep0np3sNm8HMj_e7OwoRUy0XVqD8FKYchcxLoUs?ext=json"
"https://arweave.net/kSb4SqqpGwdVVeyQ8JPhT_htIbq6Mn0gWLguPqF5gow"
"https://www.arweave.net/_n-rDupBSHsJoA9qDux-eR7KjLxaXR5TE7724EAazcw?ext=json"

rocks-db/src/columns/offchain_data.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl From<OffChainDataGrpc> for OffChainData {
+    fn from(off_chain_data: OffChainDataGrpc) -> Self {
+        let storage_mutability = StorageMutability::from(off_chain_data.url.as_str());
+        OffChainData {
+            storage_mutability,
+            url: Some(off_chain_data.url),
+            metadata: Some(off_chain_data.metadata),
+            last_read_at: 0,
```
StanChe, line 54: lets add a todo to add it to grpc if we get to populating it in p2p

rocks-db/src/lib.rs
```diff
@@ -827,3 +809,14 @@ impl Storage {
         }
     }
 }
+
+#[allow(unused_variables)]
+pub trait ToFlatbuffersConverter<'a> {
+    type Target: 'a;
+    fn convert_to_fb(&self, builder: &mut FlatBufferBuilder<'a>) -> WIPOffset<Self::Target> {
+        todo!()
```
StanChe, line 817: just out of curiosity, why do we define a default implementation that'll panic in runtime?

rocks-db/src/lib.rs
StanChe, line 1: I don't see the merge function being set for both the deprecated structure and the new one.

rocks-db/src/migrations/clean_update_authorities.rs
```diff
@@ -23,9 +24,14 @@ impl From<AssetCollectionBeforeCleanUp> for AssetCollection {
     }
 }
 
+impl<'a> ToFlatbuffersConverter<'a> for AssetCollection {
```
StanChe, line 27: So this implementation will panic on runtime, is that expected

rocks-db/src/migrations/clean_update_authorities.rs
StanChe, line 1: I'd suggest we drop all the old migrations. Those were one offs and there is no environments left where those may be applied

rocks-db/src/migrations/offchain_data.rs
```diff
@@ -0,0 +1,23 @@
...
+            metadata: Some(value.metadata),
+            last_read_at: 0,
+        }
+    }
+}
+
+pub(crate) struct OffChainDataMigration;
+impl RocksMigration for OffChainDataMigration {
+    const VERSION: u64 = 4;
+    const DESERIALIZATION_TYPE: SerializationType = SerializationType::Flatbuffers;
```
StanChe, line 19: looks like this one should be bincode.
Just checked through code. It's one of cbor (insert_gaped_data - luckily never used) or bincode (transaction client and json worker)

rocks-db/src/migrations/spl2022.rs
StanChe, line 1: drop it

StanChe (commented):
This is amazing job, thank you! Improved the alignment of the code in the packages. Proper approach without cutting corners.
I'd suggest we remove old migrations altogether as all of our DBs were started from scratch with those CompleteAssetDetails in flatbuf. Some comments are related to refinement. We should also understand how do we test the change. 
I'd suggest we take one of the small environments, copy it and run the migrations. Another approach will be creating a test that fills in some off chain data using the deprecated column and runs the migration afterwards.

rocks-db/src/clients/asset_client.rs
```diff
@@ -221,8 +218,8 @@ impl Storage {
             .into_iter()
             .filter_map(|asset| {
                 asset
-                    .filter(|a| !a.metadata.is_empty())
-                    .map(|a| (a.url.clone(), a))
+                    .filter(|a| a.url.is_some() && !a.url.clone().unwrap().is_empty())
```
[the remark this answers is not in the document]
kstepanovdev: No, I haven't noticed. Brain fog.

rocks-db/src/columns/offchain_data.rs
kstepanovdev: Thanks!

rocks-db/src/columns/offchain_data.rs
kstepanovdev: Shouldn't each peer keep it's own "last read at" info?

rocks-db/src/lib.rs
kstepanovdev: Ugly patch to preserve backward compatibility. It will now panic at runtime because atm no structures without flatbuffers call those methods. But yes, it's better to be dropped.

rocks-db/src/migrations/clean_update_authorities.rs
kstepanovdev: Wdym? It will not. `AssetCollection` doesn't do calls to flatbuff methods. In case it did, it would panic.

rocks-db/src/lib.rs
kstepanovdev: As far as I can see, it doesn't need to have a dedicated merge function. Migrations will be done via migrator, while merge insert in terms of rocks doesn't use OffChainData as a separated structure. It has the "empty" implementation because it's a part of AssetCompleteDetails. Have I gotten something wrong?

rocks-db/src/clients/asset_client.rs
```diff
@@ -221,8 +218,14 @@ impl Storage {
             .into_iter()
             .filter_map(|asset| {
                 asset
-                    .filter(|a| !a.metadata.is_empty())
-                    .map(|a| (a.url.clone(), a))
+                    .filter(|a| {
```
StanChe, line 221: Once again, we don't check the metadata object for emptiness. So we may end up with a hashmap containing a url and an empty value. Is it expected from now on?

rocks-db/src/columns/offchain_data.rs
```diff
@@ -0,0 +1,141 @@
...
+            StorageMutability::Immutable => false,
+            StorageMutability::Mutable => true,
+        }
+    }
+}
+
+impl From<&str> for StorageMutability {
+    fn from(storage_mutability: &str) -> Self {
+        if storage_mutability.is_empty()
+            || storage_mutability.starts_with("")
```
StanChe, line 28: This will match any string imaginable, right?

rocks-db/src/columns/offchain_data.rs
StanChe: It really depends on the level of trust to those peers. Indicating the last checked time over grpc when sharing that data seems a valid thing to do.

rocks-db/src/migrator.rs
```diff
@@ -67,11 +76,11 @@ impl Storage {
...
-        //     .map_err(StorageError::Common)?;
-        // let migration_applier =
-        //     MigrationApplier::new(db_path, migration_storage_path, applied_migrations);
+        let applied_migrations = migration_version_manager
+            .get_all_applied_migrations()
+            .map_err(StorageError::Common)?;
+        let migration_applier =
+            MigrationApplier::new(db_path, migration_storage_path, applied_migrations);
 
         // // apply all migrations
```
StanChe, line 85: This may be dropped as well

rocks-db/src/migrations/offchain_data.rs
```diff
@@ -0,0 +1,64 @@
...
+        Self {
+            storage_mutability: immutability,
+            url: Some(value.url),
+            metadata: Some(value.metadata),
+            last_read_at: 0,
+        }
+    }
+}
+
+impl OffChainData {
```
StanChe, line 21: Why is it implemented on the new OffChainData, while it operates on the OffChainDataDeprecated?

StanChe: And it's also putting the data back in the flatbuffer format. This seems wrong in terms of the merge function

rocks-db/src/lib.rs
StanChe: Now we have the merge function for new data column, that's operating on the old format, which seems wrong. And no merge function for the existing (deprecated) column. This code will not start with the existing db, as it requires a merge function to be present if it was defined once. 
I believe you're missing a bit context on how merge works.
the db has the deprecated column already. It may also have a list of "pending merges" for that column. So when you open the DB for the migration, the old column will be read and data in it will be merged using a merge function on that "deprecated" column. Then the data will move to the new column and whenever it'll be merged with new values a different merge function will be called for it (the one defined for the new _V2 column). it'll have the existing value in flatbuffer already, as well as the merge operands (unless you're merging in a different format - which is also possible, but will introduce a bit of confusion). The result should also be in flatbuf

StanChe (commented):
I'd suggest creating a test scenario to verify it.

rocks-db/src/columns/offchain_data.rs
kstepanovdev: Right ⚰️

StanChe (commented):
Great job. I'll add a test to verify it's working

StanChe (changes requested):
Added a test and some fixes. The test is failing - after the migration no data was copied to a new format. 
The biggest change required is that the read size was not modified and is still reading as if the data is in bincode. The example of reading flatbuf may be found in AssetComplete details. In short - any of the existing get methods may not be used.

StanChe (approved):
Great job, thank you! I'll put it live today on Eclipse

StanChe (changes requested):
Wait a moment. Any interaction with asset_offchain_data, like batch_get or get should be changed to a flatbuffed version

StanChe (approved):

PR #341, opened 2024-12-17
[MTG-1053] simplified synchronizer by splitting the nft from fungible sync

StanChe:


nft_ingester/src/bin/synchronizer/main.rs
```diff
@@ -139,64 +140,33 @@ pub async fn main() -> Result<(), IngesterError> {
...
+    
+                match result {
+                    Ok(_) => {
+                        tracing::info!("{:?} Synchronization finished successfully", asset_type)
+                    }
+                    Err(e) => tracing::error!("{:?} Synchronization failed: {:?}", asset_type, e),
+                }
+    
+                tokio::time::sleep(tokio::time::Duration::from_secs(
+                    config.timeout_between_syncs_sec,
```
kstepanovdev, line 156: shutdown will have to wait if it receives the signal.

kstepanovdev (approved):

nft_ingester/src/bin/synchronizer/main.rs
StanChe: Change to tokio::select over sleep and cancelation token. Please let me know what you think of this approach

PR #343, opened 2024-12-19
MTG-1057 Add consistency check tools

n00m4d:
# What
This PR adds new package with two binaries: compressed_assets and regular_assets. That two binaries are verifying if our RocksDB is consistent.

# How

## Compressed assets
This binary takes a csv file with cNFT tree addresses and verifying if each minted asset in a tree has valid proofs.

## Regular assets
This binary takes solana accounts snapshot, iterates over it and checks if RocksDB contains all the NFTs and token accounts.


consistency_check/src/bin/compressed_assets/main.rs
```diff
@@ -0,0 +1,431 @@
...
+                    let api_metrics_cloned = api_metrics.clone();
+                    let failed_proofs_cloned = failed_proofs.clone();
+                    let tree_cloned = tree.clone();
+                    let assets_processed_cloned = assets_processed.clone();
+                    let assets_with_failed_proofs_cloned = assets_with_failed_proofs.clone();
+                    let assets_with_missed_proofs_cloned = assets_with_missed_proofs.clone();
+                    let rate_cloned = rate.clone();
+                    let progress_bar_cloned = progress_bar.clone();
+                    tokio::spawn(async move {
+                        let asset_id = mpl_bubblegum::utils::get_asset_id(&t_key, asset_index);
```
andrii-kl, line 231: I would move internal body in to separate function and add some simple test on tree validation.

andrii-kl, line 231: get_asset_id throw  panic!("Unable to find a viable program address bump seed")
I think we should process it here to avoid interruption of the entire verification process

n00m4d: thats barely happen. because we generate asset ids with indexes which were already used in key generation, such as we take supply number from tree config first and then generate asset id for each index.

andrii-kl (approved):

consistency_check/src/bin/regular_assets/main.rs
```diff
@@ -0,0 +1,427 @@
...
+        Pubkey::from_str("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA").unwrap();
+    static ref SPL_2022_KEY: Pubkey =
+        Pubkey::from_str("TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb").unwrap();
+}
+
+const MINT_ACC_DATA_SIZE: usize = 82;
+
+const CHANNEL_SIZE: usize = 100_000_000;
+
+const MISSED_ASSET_FILE: &str = "./missed_asset_data.csv";
```
StanChe, line 56: Should be an arg with PathBuf type and this value as default

rocks-db/src/column.rs
```diff
@@ -512,4 +512,13 @@ where
         let res = db.get_cf(cf, &encoded_key)?;
         Ok(res.is_some())
     }
+
+    pub async fn key_exist(&self, key: C::KeyType) -> Result<bool> {
+        let db = self.backend.clone();
+        tokio::task::spawn_blocking(move || {
+            let cf = &db.cf_handle(C::NAME).unwrap();
+            let encoded_key = C::encode_key(key);
+            Ok(db.key_may_exist_cf(cf, &encoded_key))
```
StanChe, line 521: This may produce false positives. We already have the method 
```
  pub(crate) fn sync_has_key(db: Arc<DB>, key: C::KeyType) -> crate::Result<bool> {
        let cf = &db.cf_handle(C::NAME).unwrap();
        let encoded_key = C::encode_key(key);
        if !db.key_may_exist_cf(cf, &encoded_key) {
            return Ok(false);
        }
        let res = db.get_cf(cf, &encoded_key)?;
        Ok(res.is_some())
    }
```
I'd suggest wrapping it with async instead

StanChe (commented):
Mostly spelling, several comments on args types and some improvements suggested. It'll work as it is, so either create additional tickets and merge or fix some comments and put non done into tickets

consistency_check/src/bin/regular_assets/snapshot_reader.rs
StanChe, line 1: this looks like a mix of some custom code and some code copied from Solana. If anything changes on their end, it'll be tricky to support the changes. I presume we couldn't reuse the types from Solana, could we? If we couldn't I'd suggest putting all the copied code in a separate file/files with minimal to no changes and adding custom logic beside it.

n00m4d: I took that file from [this repo](https://github.com/metaplex-foundation/digital-asset-validator-plugin/blob/main/plerkle_snapshot/src/solana.rs). Agree that it would be better to have it in one place at least but it also requires some refactor on digital asset validator repos side. Will create a ticket not to forget about this refactor, thanks.

StanChe (approved):
Great, thank you!

consistency_check/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+) {
+    let mut last_time = std::time::Instant::now();
+    let mut last_count = assets_processed.load(Ordering::Relaxed);
+
+    loop {
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
+        if shutdown_token.is_cancelled() {
+            break;
+        }
```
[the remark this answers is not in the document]
StanChe: @n00m4d you'll need to fix the missing closing bracket

PR #345, opened 2024-12-26
[MTG-1065] encode/decode as part of the column specification

StanChe:


kstepanovdev (approved):
Looks amazing, thank you!

PR #349, opened 2025-01-06
MTG-1110 Add request time out for DB query

andrii-kl:
- Added a request timeout configuration for database queries.

This change allows setting a maximum duration for database queries to prevent long-running operations from blocking the system.


andrii-kl:
implemented another solution https://github.com/metaplex-foundation/aura/pull/362

PR #350, opened 2025-01-07
fix: make tests work

armyhaylenko:


rocks-db/tests/dump_tests.rs
```diff
@@ -59,7 +59,6 @@ async fn test_scv_export_from_rocks() {
         .has_headers(false)
         .from_path(creators_path)
         .unwrap();
-    assert_eq!(tasks_reader.records().count(), 1);
```
StanChe, line 62: this should still be valid - tasks is an old name of metadata_file

rocks-db/tests/batch_client_integration_tests.rs
```diff
@@ -350,117 +350,4 @@ mod tests {
...
-        let mut dynamic_data = create_test_dynamic_data(pk, 0, "http://example.com".to_string());
-        dynamic_data.is_compressible = Updated::new(0, None, false);
-        dynamic_data.is_compressed = Updated::new(0, None, false);
-        dynamic_data.supply = Some(Updated::new(0, None, 1));
-
-        storage
-            .asset_dynamic_data
-            .merge(dynamic_data.pubkey, dynamic_data.clone())
-            .await
-            .unwrap();
```
StanChe, line 367: It would be great if we could maintain the logic of the test. a todo with a commented out + a ticket, or even moving the test into a linear ticket will be ok. A little background: 
The underlying structures were modified completely, and in current version look more like (taken from rocks-db/src/transaction_client.rs)

```
let acd = AssetCompleteDetails {
  pubkey: pk,
  static_details: None,
  dynamic_details: update.update.as_ref().and_then(|u| u.dynamic_data.clone()),
  owner: None,
  authority: None,
  collection: None,
};
let mut builder = flatbuffers::FlatBufferBuilder::with_capacity(2500);
let acd = acd.convert_to_fb(&mut builder);
builder.finish_minimal(acd);
batch.merge_cf(
  &self.db.cf_handle(AssetCompleteDetails::NAME).unwrap(),
  pk,
  builder.finished_data(),
);
```

StanChe (commented):
Great job, thank you! Just couple comments

rocks-db/tests/batch_client_integration_tests.rs
armyhaylenko: i will create a ticket specifically for these two test cases, move their code there, but i still suggest removing this code from the repo

PR #359, opened 2025-01-13
[mtg-1162] URI trimming

kstepanovdev:
Stop trimming whitespace in the name and metadata of assets to prevent invalid indexing.


armyhaylenko (approved):
LBTM

PR #363, opened 2025-01-15
[mtg-808] optimize select for show fungible display option

kstepanovdev:
The PR optimizes the search request via dropping OR condition if a search for all token types is requested. OR condition in such case provokes seq scan through the table. 
To avoid this, UNION between two selects is used instead.

### DIFF
--- a/postgre-client/src/asset_filter_client.rs
+++ b/postgre-client/src/asset_filter_client.rs
@@ -35,7 +35,7 @@ impl PgClient {
         options: &'a GetByMethodsOptions,
     ) -> Result<(QueryBuilder<'a, Postgres>, bool), IndexDbError> {
         let mut query_builder = QueryBuilder::new(
-            "SELECT ast_pubkey pubkey, ast_slot_created slot_created, ast_slot_updated slot_updated FROM assets_v3 ",
+            "(SELECT ast_pubkey pubkey, ast_slot_created slot_created, ast_slot_updated slot_updated FROM assets_v3 ",
         );
         let group_clause_required = add_filter_clause(&mut query_builder, filter, options);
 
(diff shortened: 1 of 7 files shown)


postgre-client/src/asset_filter_client.rs
```diff
@@ -103,19 +103,53 @@ impl PgClient {
...
+        // the function with a side-effect that mutates the query_builder
+        if let Some(owner_address) = &filter.owner_address {
+            if let Some(ref token_type) = filter.token_type {
+                match *token_type {
+                    TokenType::Fungible
+                    | TokenType::NonFungible
+                    | TokenType::RegularNFT
+                    | TokenType::CompressedNFT => {}
+                    TokenType::All => {
+                        query_builder.push("  UNION ALL ");
```
StanChe, line 119: Do you need `union all` here? It might return you duplicate asset_ids, which doesn't look like expected behaviour. Although to be fair it should not receive duplicates in this union, I'd suggest using just the `UNION ` here

postgre-client/src/asset_filter_client.rs
```diff
@@ -262,14 +296,11 @@ fn add_filter_clause<'a>(
                 TokenType::All => {
                     query_builder.push(" AND (assets_v3.ast_owner = ");
                     query_builder.push_bind(owner_address);
-                    query_builder.push(" OR (fungible_tokens.fbt_owner = ");
-                    query_builder.push_bind(owner_address);
```
StanChe, line 266: looks like the below `if` should also be dropped. As well as the whole join on fungible_tokens. Also @n00m4d might be working on a concurrent task with balances. I believe with your change and dropped join Vadim task will also be solved.

postgre-client/src/asset_filter_client.rs
kstepanovdev: I do not.

n00m4d (approved):

PR #364, opened 2025-01-15
MTG-1044: add a separate secondary rocks backup service

armyhaylenko:


andrii-kl (approved):

n00m4d (approved):
LGTM
Let's also add command to the Makefile

docker-compose.yaml
```diff
@@ -161,3 +161,25 @@ services:
     logging:
       options:
         max-size: "100m"
+
+  rocksdb-backup:
+    container_name: rocksdb-backup
+    restart: always
```
StanChe, line 167: It's a one off, why do we need to restart always?

armyhaylenko: it works based on an interval that we specify using env. in this case, it's true that it will immediately retrigger after restart, but shouldn't this be the intended behavior?

StanChe: Got it, it's working continuously atm. Wdyt of making it a one off - run it when you need it. When done - exit(0). As for me that's way easier to integrate on any ops side. If we run it continuously - we have to track when the backup is complete. If we stop it when it started the backup - we'll end up without a valid backup and so on.

docker-compose.yaml
```diff
@@ -161,3 +161,23 @@ services:
...
+    container_name: rocksdb-backup
+    entrypoint: ./rocksdb_backup
+    env_file:
+      - .env
+    network_mode: host
+    volumes:
+      - ${ROCKS_DB_PATH}:${ROCKS_DB_PATH_CONTAINER}:ro
+      - ${ROCKS_BACKUP_DIR}:${ROCKS_BACKUP_DIR}:rw
+      - ${ROCKS_BACKUP_ARCHIVES_DIR}:${ROCKS_BACKUP_ARCHIVES_DIR}:rw
+      - ${ROCKS_DB_SECONDARY_PATH}:${ROCKS_DB_SECONDARY_PATH_CONTAINER}:rw
```
StanChe, line 175:
```suggestion
      - ${ROCKS_DB_SECONDARY_PATH}/backup:${ROCKS_DB_SECONDARY_PATH_CONTAINER}:rw
```

nft_ingester/src/config.rs
```diff
@@ -306,6 +306,31 @@ pub struct SynchronizerClapArgs {
     pub log_level: String,
 }
 
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+pub struct RocksDbBackupServiceClapArgs {
+    #[clap(long, env, default_value = "./my_rocksdb", help = "Rocks db path container")]
+    pub rocks_db_path_container: String,
```
StanChe, line 313: may be PathBuf, as well as the other paths below

nft_ingester/src/config.rs
```diff
@@ -306,6 +306,31 @@ pub struct SynchronizerClapArgs {
...
+    pub backup_dir: String,
+    #[clap(
+        long,
+        env = "ROCKS_FLUSH_BEFORE_BACKUP",
+        help = "Whether to flush RocksDb before backup"
+    )]
+    pub flush_before_backup: bool,
+
+    #[clap(long, env, default_value = "/usr/src/app/heaps", help = "Heap path")]
+    pub heap_path: String,
```
StanChe, line 328: It doesn't look as a required argument, and it's not used afai can see

nft_ingester/src/bin/rocksdb_backup/main.rs
```diff
@@ -0,0 +1,59 @@
...
+    let mut backup_service = RocksDbBackupService::new(
+        rocks_storage.db.clone(),
+        &RocksDbBackupServiceConfig {
+            rocks_backup_dir: args.backup_dir,
+            rocks_backup_archives_dir: args.backup_archives_dir,
+            rocks_flush_before_backup: args.flush_before_backup,
+        },
+    )?;
+
+    mutexed_tasks.lock().await.spawn(async move { backup_service.perform_backup().await });
```
StanChe, line 56: You're spawning, but not joining this task - will it reach its completion?

armyhaylenko: it does but very good point, it's a tiny memory leak

StanChe (approved):
Great job, thank you!

PR #365, opened 2025-01-15
MTG-1155: Fix Redis processing messages more than once

andrii-kl:
- Add a unique consumer ID for each worker.
- Add a unique worker name to simplify debugging.

### DIFF
--- a/nft_ingester/src/config.rs
+++ b/nft_ingester/src/config.rs
@@ -44,17 +44,20 @@ pub struct IngesterClapArgs {
     #[clap(short('r'), long, env, help="example: {redis_connection_str=\"redis://127.0.0.1:6379/0\"}", value_parser = parse_json_to_dict)]
     pub redis_connection_config: Dict,
 
-    #[clap(long, env, default_value = "20")]
+    #[clap(long, env, default_value = "5")]
     pub redis_accounts_parsing_workers: u32,
 
-    #[clap(long, env, default_value = "20")]
+    #[clap(long, env, default_value = "1")]
     pub redis_transactions_parsing_workers: u32,
 
     #[clap(long, env, default_value = "60")]
     pub sequence_consistent_checker_wait_period_sec: u64,
 
-    #[clap(long, env, default_value = "250")]
+    #[clap(long, env, default_value = "100")]
     pub account_processor_buffer_size: usize,
+
+    #[clap(long, env, default_value = "100")]
+    pub tx_processor_buffer_size: usize,
 
     #[clap(long, env, default_value = "50")]
     pub account_processor_mpl_fees_buffer_size: usize,
--- a/nft_ingester/src/processors/accounts_processor.rs
+++ b/nft_ingester/src/processors/accounts_processor.rs
@@ -21,6 +21,7 @@ use tokio::{
     time::Instant,
 };
 use tracing::{debug, error};
+use uuid::Uuid;
 
 use super::account_based::{
     inscriptions_processor::InscriptionsProcessor,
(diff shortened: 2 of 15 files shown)


nft_ingester/src/processors/accounts_processor.rs
```diff
@@ -163,6 +168,13 @@ impl<T: UnprocessedAccountsGetter> AccountsProcessor<T> {
                                 continue;
                             }
                         };
+
+                        if unprocessed_accounts.is_empty() {
+                            tokio::time::sleep(Duration::from_millis(1000)).await;
```
StanChe, line 173: Why do we sleep here for so long? 400ms is the time of a slot, so we'll have 2.5 slots closed and a lot of messages after the sleep. I'd suggest smth <100ms might be tolarable. Currently, when there are no congestions and not much interactions we would see the following p90 latency from redis receiving the message to it being processed: 
image
so even 50ms will double this metric

andrii-kl: In the situation when we have more than one worker and there is a batch size for example 100, then the worker will select these 100 blocks and the rest will not have a message, therefore when there are too few messages in the queue, a large number of employees does not make sense. But in general we need to see with different settings how the workers will handle messages and which setting will be better for our system.

nft_ingester/src/config.rs
```diff
@@ -560,11 +563,7 @@ mod tests {
         assert_eq!(args.rocks_db_path_container, "./my_rocksdb");
         assert_eq!(args.file_storage_path_container, "./tmp/file_storage");
         assert_eq!(args.pg_max_db_connections, 100);
-        assert_eq!(args.redis_accounts_parsing_workers, 20);
-        assert_eq!(args.redis_transactions_parsing_workers, 20);
         assert_eq!(args.sequence_consistent_checker_wait_period_sec, 60);
-        assert_eq!(args.account_processor_buffer_size, 250);
-        assert_eq!(args.account_processor_mpl_fees_buffer_size, 50);
```
armyhaylenko, line 567: why remove these assertions and not replace with the same changes done to IngesterClapArgs?

nft_ingester/src/processors/accounts_processor.rs
```diff
@@ -163,6 +168,9 @@ impl<T: UnprocessedAccountsGetter> AccountsProcessor<T> {
                                 continue;
                             }
                         };
+
+                        debug!("Processor {}, Unprocessed_accounts: {}  {:?}", self.processor_name, unprocessed_accounts.len(), unprocessed_accounts.iter().map(|account| account.id.to_string()).collect::<Vec<_>>().join(", "));
```
armyhaylenko, line 172: as discussed internally, please log the values as follows:
```suggestion
                        debug!(processor = %self.processor_name, unprocessed_accounts_len = %unprocessed_accounts.len(), unprocessed_accounts = ?unprocessed_accounts.iter().map(|account| account.id.to_string()).collect::<Vec<_>>().join(", "), "Processor {}, Unprocessed_accounts: {}  {:?}", self.processor_name, unprocessed_accounts.len(), unprocessed_accounts.iter().map(|account| account.id.to_string()).collect::<Vec<_>>().join(", "));
```

feel free to optimize this as to not iterate twice on unprocessed accounts, but logging the values outside of the "message"/"msg" field in the JSON is a must.

nft_ingester/src/processors/accounts_processor.rs
```diff
@@ -125,6 +127,7 @@ impl<T: UnprocessedAccountsGetter> AccountsProcessor<T> {
             MplCoreFeeProcessor::build(postgre_client, metrics.clone(), rpc_client, join_set)
                 .await?;
         core_fees_processor.update_rent(rx).await;
+        let processor_name = Uuid::new_v4().to_string();
```
StanChe, line 130: I'd suggest we use the same uuid for redis and for the processor. Either inject it into both or at least in one of those

StanChe (approved):

n00m4d (approved):

nft_ingester/src/config.rs
andrii-kl: Not important parameters, It takes a long time to maintain positive tests. 
the main goal is check that the specified default parameters are taken

PR #368, opened 2025-01-16
MTG 1164 Apply showZeroBalance only for NFTs

n00m4d:
# What

This PR adds condition to the JOIN SQL command which tells to join only rows with `ast_owner_type` field set to `token`. As a result from `fungible_tokens` table will be taken accounts which are related to fungible assets and not NFTs.

Also this PR adds new integration test which checks if that JOIN works as expected.


armyhaylenko (approved):
LBTM

postgre-client/src/asset_filter_client.rs
```diff
@@ -123,7 +123,7 @@ impl PgClient {
                     ast_slot_created,
                     ast_slot_updated
                     FROM assets_v3
-                    JOIN fungible_tokens ON ast_pubkey = fungible_tokens.fbt_asset
+                    JOIN fungible_tokens ON ast_pubkey = fungible_tokens.fbt_asset AND ast_owner_type = 'token'
```
andrii-kl, line 126: Can there be an asset with **ast_owner_type** **single** or **unknown**? (because as far as I understand we apply this change to all possible requests)

n00m4d: This PR https://github.com/metaplex-foundation/aura/pull/376 adds fix not to have `unknown` in `ast_owner_type` anymore.
And we join only rows with `ast_owner_type = 'token'` because for NFTs(rows with `ast_owner_type = 'single'`) there is no necessary information in `fungible_tokens` table.

n00m4d: If we join rows with `ast_owner_type='single'` we start to return NFTs with zero balance, which was identified as unwanted behaviour.

andrii-kl (approved):

postgre-client/src/asset_filter_client.rs
StanChe: How are the indexes being used for this join?

StanChe (approved):
Looks great, thank you! The only question is whether we need to modify any indexes

n00m4d:
@StanChe nope, we should not. I've tested this query on nodes with Solana data and with that added condition it works same as without it. We are good there.

n00m4d:
Double checked explain analyze for final query, these changes actually increases execution time, from >100ms to ~3 seconds. First tests I did showed incorrect results. So it's better not to merge this PR at the moment.

n00m4d:
Latest commit adds improvements to SQL query. Also it adds 2 new indexes which also will improve selects.
Now it's ready for final review.

migrations/12_fungible_and_assets_indexes.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX fungible_tokens_fbt_owner_fbt_asset_fbt_balance_idx ON fungible_tokens (fbt_owner, fbt_asset, fbt_balance);
+
+CREATE INDEX assets_ast_owner_type_ast_pubkey_idx ON assets_v3 (ast_owner_type, ast_pubkey);
```
StanChe, line 3: new line please

StanChe, line 3: you may also want to drop assets_v3_owner_type as it becomes obsolete

postgre-client/src/asset_filter_client.rs
```diff
@@ -137,18 +162,29 @@ impl PgClient {
...
                             );
                         }
+
+                        query_builder.push(" ORDER BY ");
+                        query_builder.push(order.sort_by.to_string());
+                        query_builder.push(direction);
+
+                        query_builder.push(" LIMIT ");
+                        if let Some(page_num) = page.filter(|&p| p > 1) {
+                            let lim = (page_num - 1) * limit;
```
StanChe, line 172: The condition is wrong. it's page_num*limit. You want to have 2000 records pre filtered for the second page. Also add a test for the second page of pagination of this request. The correct way to do it is you always get a 1-based page and multiply it by the limit

postgre-client/src/asset_filter_client.rs
```diff
@@ -113,17 +118,37 @@ impl PgClient {
...
+                        // For this type of query we do union and apply limit with ordering for both parts of a query.
+                        // It allows us to speed up the query for queries with wallets which has lots of assets.
+                        // Not the cleanest approach.
+                        // TODO: this may be improved
+                        query_builder.push(" ORDER BY ");
+                        query_builder.push(order.sort_by.to_string().replace("ast_", ""));
+                        query_builder.push(direction);
+
+                        query_builder.push(" LIMIT ");
+                        if let Some(page_num) = page.filter(|&p| p > 1) {
```
StanChe, line 134: same as below

StanChe (changes requested):
Great change, let's ask others to revalidate as well. The limit should be fixed

StanChe (approved):

integration_tests/src/regular_nft_tests.rs
```diff
@@ -250,6 +250,98 @@ async fn test_search_by_owner_with_show_zero_balance() {
     insta::assert_json_snapshot!(name, response);
 }
 
+#[tokio::test]
+#[serial]
+#[named]
+async fn test_requested_non_fungibles_are_non_fungibles_and_page() {
```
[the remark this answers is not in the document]
n00m4d: Thanks. Renamed test itself, and added one more case to the test with select all. Also don't see the point write any additional sequence check if we already saved and verified that snapshots are correct. They will not be changed.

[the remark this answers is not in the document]
andrii-kl: in the current solution you need to manually check the snapshots, but in the one I propose this check would be automated. (not critical)

PR #371, opened 2025-01-16
MTG-1110 Add request timeout for PG DB query

andrii-kl:
- Added configurable timeout for the PG database queries
- Added environment variables to configure PG database timeouts

### DIFF
--- a/nft_ingester/src/api/dapi/search_assets.rs
+++ b/nft_ingester/src/api/dapi/search_assets.rs
@@ -148,7 +148,14 @@ async fn fetch_assets<
     let keys = index_client
         .get_asset_pubkeys_filtered(filter, &sort_by.into(), limit, page, before, after, &options)
         .await
-        .map_err(|e| StorageError::Common(e.to_string()))?;
+        .map_err(|e| {
+            if e.to_string().contains("statement timeout") {
+                StorageError::QueryTimedOut
+            } else {
+                StorageError::Common(e.to_string())
+            }
+        })?;
+
     let asset_ids = keys
         .iter()
         .filter_map(|k| Pubkey::try_from(k.pubkey.clone()).ok())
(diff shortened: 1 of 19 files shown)


nft_ingester/src/api/error.rs
```diff
@@ -95,6 +97,11 @@ impl From<DasApiError> for jsonrpc_core::Error {
                     .to_string(),
                 data: None,
             },
+            DasApiError::QueryTimedOut => jsonrpc_core::Error {
+                code: ErrorCode::InvalidRequest,
```
StanChe, line 101: Probably it's more of an InternalError, or rather ServerError with some custom code. At least I've seen from some other providers
```
"error": {
    "code": -32504,
    "message": "Request timed out, please try again."
  }
```

StanChe (approved):

nft_ingester/src/api/error.rs
andrii-kl: I see this error more as a badly written query that takes longer than it should. Therefore, in my opinion, this error is not related to our server, but to the specifics of the client’s request, who wrote an inappropriate request.

nft_ingester/src/api/dapi/search_assets.rs
```diff
@@ -148,7 +148,14 @@ async fn fetch_assets<
...
         .await
-        .map_err(|e| StorageError::Common(e.to_string()))?;
+        .map_err(|e| {
+            if e.to_string().contains("statement timeout") {
+                StorageError::QueryTimedOut
+            } else {
+                StorageError::Common(e.to_string())
+            }
+        })?;
+
```
armyhaylenko, line 158: would be cool if we could statically type the error in IndexDbError, but if it's not possible then i'm ok with what written here and below

nft_ingester/src/init.rs
```diff
@@ -24,12 +24,19 @@ pub async fn init_index_storage_with_migration(
...
+        url,
+        min_pg_connections,
+        max_pg_connections,
+        base_dump_path,
+        red_metrics,
+        pg_max_query_statement_timeout_secs,
+    )
+    .await
+    .map_err(|e| e.to_string())
+    .map_err(IngesterError::SqlxError)?;
```
armyhaylenko, line 39:
```suggestion
    .map_err(|e| IngesterError::SqlxError(e.to_string()))?;
```

armyhaylenko (approved):
LBTM, feel free to add suggested changes

n00m4d (approved):

StanChe:
currently we have
```
Error: SqlxError("error returned from database: unsupported startup parameter in options: statement_timeout")
```

PR #376, opened 2025-01-20
MTG 1195 Merge function update

n00m4d:
# What
This PR slightly changes merge function for AssetStaticDetails. The problem was that owner_type row could be rewrited with old value from other_owners vec without mergin, and such as in other_owners vec mostly stored only updates from token accounts owner type value there Default.

As a result in PG table assets_v3 row ast_owner_type we have `unknown`, this field can affect different API requests.

Also in this PR modified integration test which shows that even if we change sequence of updates received from Geyser the final state in the DB will not be affected by it.

# Release steps
Such as this PR changes merge function data in Rocks should be reindexed. Taking into account that it affects only account based protocol parsing new snapshot with slot higher that saved in the DB will be enough.


armyhaylenko (approved):
LBTM

rocks-db/src/columns/asset.rs
```diff
@@ -2073,12 +2071,18 @@ pub fn merge_complete_details_fb_simple_raw<'a>(
...
                                 owner_owner = merged_owner.owner;
                                 owner_delegate = merged_owner.delegate;
-                                owner_owner_type = merged_owner.owner_type;
                                 owner_owner_delegate_seq = merged_owner.owner_delegate_seq;
                                 owner_is_current_owner = merged_owner.is_current_owner;
                                 other_known_owners.remove(new_owner_pubkey.bytes());
                             }
                         }
+                        // Since the owner type can only be changed through a mint account state update,
+                        // we should avoid modifying it every time we receive token account updates for the corresponding mint.
```
andrii-kl, line 2080: won't there be an issue when solving the fork if the owner_type on the alternative chain is different?

n00m4d: it's hard to imagine such a scenario because owner_type is changing only when token supply is changing, and it's not happening that often as token transfers, and it will be changed only when supply is changing from `1` to `>1`. also we don't have yet any proofs that forks affect accounts same as transactions. and if it is then we should rethink merge logic for all the fields, not only owner_type, because at the moment we simply merge everything.

that's a really good point, thanks for it. but I think we need to have proofs first to understand what exactly happening there and than refactor code if it needs.

andrii-kl (approved):

StanChe (approved):

PR #378, opened 2025-01-22
MTG 995 Add integration tests with full dbs sync scenario

n00m4d:
# What

This PR adds integration tests with full databases synchronization scenario.

Also there is additional bash script to launch Postgre for tests was added.


StanChe (approved):
Great, thank you! Having more tests is always great. As a side note - now we have 2 setups to run integration tests with postgreSQL

PR #379, opened 2025-01-22
MTG-1242 Fix build search query for Fungible tokens. MTG-1031 Fix tes…

andrii-kl:
…t for fungible tokens.

### DIFF
--- a/nft_ingester/src/api/dapi/search_assets.rs
+++ b/nft_ingester/src/api/dapi/search_assets.rs
@@ -36,7 +36,7 @@ pub async fn search_assets<
 >(
     index_client: Arc<impl postgre_client::storage_traits::AssetPubkeyFilteredFetcher>,
     rocks_db: Arc<Storage>,
-    filter: SearchAssetsQuery,
+    mut filter: SearchAssetsQuery,
     sort_by: AssetSorting,
     limit: u64,
     page: Option<u64>,
@@ -55,6 +55,9 @@ pub async fn search_assets<
     tree_gaps_checker: &Option<Arc<PPC>>,
     native_mint_pubkey: &str,
 ) -> Result<AssetList, StorageError> {
+    if options.show_fungible {
+        filter.token_type = Some(TokenType::All)
+    }
     let show_native_balance = options.show_native_balance;
     let (asset_list, native_balance) = tokio::join!(
         fetch_assets(
--- a/nft_ingester/src/api/dapi/search_assets.rs
+++ b/nft_ingester/src/api/dapi/search_assets.rs
@@ -36,7 +36,7 @@ pub async fn search_assets<
 >(
     index_client: Arc<impl postgre_client::storage_traits::AssetPubkeyFilteredFetcher>,
     rocks_db: Arc<Storage>,
-    mut filter: SearchAssetsQuery,
+    filter: SearchAssetsQuery,
     sort_by: AssetSorting,
     limit: u64,
     page: Option<u64>,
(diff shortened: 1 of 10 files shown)


nft_ingester/src/index_syncronizer.rs
```diff
@@ -199,9 +199,13 @@ where
                 self.regular_nft_syncronize(rx, state.last_indexed_key, state.last_known_key).await
             },
             SyncStatus::RegularSyncRequired(state) => {
+                tracing::info!("Regular sync required for nft asset");
```
StanChe, line 202: Do you really want to see several 100s of this infos per second on Eclipse?

postgre-client/src/asset_filter_client.rs
```diff
@@ -209,6 +209,21 @@ fn add_filter_clause<'a>(
         group_clause_required = true;
     }
 
+    if let Some(ref token_type) = filter.token_type {
+        if token_type == &TokenType::All && filter.owner_address.is_some() {
+            query_builder.push(
+                " LEFT JOIN fungible_tokens ON assets_v3.ast_pubkey = fungible_tokens.fbt_asset ",
```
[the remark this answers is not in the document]
kstepanovdev: Also, the if statement with zero balance may need to be applied, pls take a look at the "union" case and/or the previous version of the file that had been written before I committed "union".

kstepanovdev (changes requested):
lgtm except for the part with the joining fungible tokens, left a comment

postgre-client/src/asset_filter_client.rs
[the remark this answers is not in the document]
andrii-kl: Remove Join for TokenType::All

kstepanovdev (approved):
LGTM

nft_ingester/src/api/dapi/search_assets.rs
```diff
@@ -55,9 +55,6 @@ pub async fn search_assets<
     tree_gaps_checker: &Option<Arc<PPC>>,
     native_mint_pubkey: &str,
 ) -> Result<AssetList, StorageError> {
-    if options.show_fungible {
```
StanChe, line 58: wdyt of instead of removing this if to make the condition more precise:
```
if options.show_fungible &&  filter.token_type != Some(TokenType::Fungible) {
```
This will ensure we don't overwrite the request when the user asks specifically for fungibles

andrii-kl: This action looks very strange and not obvious, since in the middle of the function execution we change the user request option.

andrii-kl: in the current implementation this option _token_type_ / _show_fungible_ can be specified in the request, and I don't quite understand why do we need to change it in the process.

nft_ingester/tests/api_tests.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(test)]
-#[cfg(feature = "integration_tests")]
+// #[cfg(feature = "integration_tests")]
```
StanChe, line 2:
```suggestion
#[cfg(feature = "integration_tests")]
```

nft_ingester/src/index_syncronizer.rs
```diff
@@ -195,13 +195,17 @@ where
         let state = self.get_sync_state(run_full_sync_threshold, asset_type).await?;
         match state {
             SyncStatus::FullSyncRequired(state) => {
-                tracing::info!("Should run dump synchronizer as the difference between last indexed and last known sequence is greater than the threshold. Last indexed: {:?}, Last known: {}", state.last_indexed_key.clone().map(|k|k.seq), state.last_known_key.seq);
+                tracing::debug!("Should run dump synchronizer as the difference between last indexed and last known sequence is greater than the threshold. Last indexed: {:?}, Last known: {}", state.last_indexed_key.clone().map(|k|k.seq), state.last_known_key.seq);
```
StanChe, line 198: info or warn

StanChe (approved):

nft_ingester/src/index_syncronizer.rs
StanChe: SyncStatus::FullSyncRequired indicates the synchronization gap is huge and most likely synchronizer will not be able to catch up

PR #381, opened 2025-01-23
Fix/mtg-1223 improve clap implementation for Injester

andrii-kl:


nft_ingester/src/config.rs
```diff
@@ -213,13 +214,12 @@ pub struct IngesterClapArgs {
     pub big_table_config: Option<BigTableConfig>,
 
     #[clap(
-        long("disable-bubblegum-backfiller"),
-        action = ArgAction::SetFalse,
+        long,
         env = "RUN_BUBBLEGUM_BACKFILLER",
-        default_value_t = true,
-        help = "#bubbl Disable bubblegum backfiller (default: true)"
+        default_value = "true",
```
kstepanovdev, line 219: Hm. How `default_value` can be `true` if it is `Option`? Shouldn't it be then Some(true)? Do we need an option?

StanChe (approved):

nft_ingester/src/config.rs
andrii-kl: Using Option is only one case to set default bool value in true. So it's just a workaround to solve a Clap issue

PR #382, opened 2025-01-23
MTG 1235 Extend range of interfaces for supply object display

n00m4d:
# What

This PR extends range of interfaces for which we display `supply` object on API response. Basicaly there were added more interfaces which may have edition information which could be put into `supply` object.

From the changes in snapshots of integration tests we can see what is changing for the user.


StanChe (approved):
Nice, thanks!

PR #384, opened 2025-01-24
[MTG-1258] an option to turn off the json re-download, enabled by default

StanChe:
As a short-term solution for longer requests we disable the cache invalidation alongside the request.
Also we separate the inline json download timeout from the main workers timeout, letting workers wait for 30 seconds for the response, while the inline download is restricted to 3 seconds.

### DIFF
--- a/interface/src/json.rs
+++ b/interface/src/json.rs
@@ -14,6 +14,7 @@ pub trait JsonDownloader {
         &self,
         url: String,
     ) -> Result<JsonDownloadResult, crate::error::JsonDownloaderError>;
+    fn skips_refreshes(&self) -> bool;
 }
 
 #[automock]
--- a/interface/src/json.rs
+++ b/interface/src/json.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use async_trait::async_trait;
 use mockall::automock;
 
@@ -13,6 +15,7 @@ pub trait JsonDownloader {
     async fn download_file(
         &self,
         url: String,
+        timeout: Duration,
     ) -> Result<JsonDownloadResult, crate::error::JsonDownloaderError>;
     fn skips_refreshes(&self) -> bool;
 }
--- a/interface/src/json.rs
+++ b/interface/src/json.rs
@@ -17,7 +17,7 @@ pub trait JsonDownloader {
         url: String,
         timeout: Duration,
     ) -> Result<JsonDownloadResult, crate::error::JsonDownloaderError>;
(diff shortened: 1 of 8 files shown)


interface/src/json.rs
```diff
@@ -13,7 +15,9 @@ pub trait JsonDownloader {
     async fn download_file(
         &self,
         url: String,
+        timeout: Duration,
     ) -> Result<JsonDownloadResult, crate::error::JsonDownloaderError>;
+    fn skips_refreshes(&self) -> bool;
```
armyhaylenko, line 20: nit: i'd rename to `fn skip_refresh(&self) -> bool`.

armyhaylenko (approved):
LBTM

nft_ingester/src/json_worker.rs
```diff
@@ -341,6 +346,10 @@ impl JsonDownloader for JsonWorker {
             },
         }
     }
+
+    fn skips_refreshes(&self) -> bool {
```
kstepanovdev, line 350: I do not think it requires a dedicated getter. Why can't we call `self.should_skip_refreshes` ?

kstepanovdev: Nvm, in case it is requested because a caller is a generic then it's acceptable

kstepanovdev (approved):

nft_ingester/src/json_worker.rs
```diff
@@ -27,12 +27,13 @@ use crate::api::dapi::rpc_asset_convertors::parse_files;
 pub const JSON_BATCH: usize = 300;
 pub const WIPE_PERIOD_SEC: u64 = 60;
 pub const SLEEP_TIME: u64 = 1;
-pub const CLIENT_TIMEOUT: u64 = 5;
+pub const CLIENT_TIMEOUT: Duration = Duration::from_secs(30);
```
andrii-kl, line 30: maybe it makes sense to make this variable configurable throw the env?

nft_ingester/src/api/dapi/asset.rs
```diff
@@ -267,7 +269,8 @@ pub async fn get_by_ids<
                     let json_downloader = json_downloader.clone();
```
andrii-kl, line 269: What if instead of skipping the update we just make the whole update process as an asynchronous process in the background.

StanChe: That'll be a bigger story with the refresh queue. This one is a rather hot fix to deploy today and make live on Monday

PR #385, opened 2025-01-24
feat(ingester): add metrics for `RedisReceiver`

armyhaylenko:
NOTE! This PR also adds filename and line number printing to all logs emitted using `tracing::error!/info!/debug!/trace!`. This is in line with the configuration of the `log` crate we use, but nevertheless important to mention here.

The rest of the PR adds metrics describing the lifecycle of `RedisReceiver`.

### DIFF
--- a/metrics_utils/src/lib.rs
+++ b/metrics_utils/src/lib.rs
@@ -56,6 +56,7 @@ pub struct MetricState {
     pub fork_cleaner_metrics: Arc<ForkCleanerMetricsConfig>,
     pub batch_mint_processor_metrics: Arc<BatchMintProcessorMetricsConfig>,
     pub batch_mint_persisting_metrics: Arc<BatchMintPersisterMetricsConfig>,
+    pub redis_receiver_metrics: Arc<RedisReceiverMetricsConfig>,
     pub registry: Registry,
 }
 
@@ -83,6 +84,7 @@ impl MetricState {
             batch_mint_processor_metrics: Arc::new(BatchMintProcessorMetricsConfig::new()),
             batch_mint_persisting_metrics: Arc::new(BatchMintPersisterMetricsConfig::new()),
             red_metrics: Arc::new(RequestErrorDurationMetrics::new()),
+            redis_receiver_metrics: Arc::new(RedisReceiverMetricsConfig::new()),
             registry: Registry::default(),
         }
     }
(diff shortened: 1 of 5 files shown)


metrics_utils/src/lib.rs
```diff
@@ -1267,3 +1270,98 @@ impl BatchMintPersisterMetricsConfig {
...
+}
+
+impl Default for RedisReceiverMetricsConfig {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl RedisReceiverMetricsConfig {
+    pub fn new() -> Self {
```
kstepanovdev, line 1292: Why do we need `new()` that accepts zero parameters and is a full copy of `Default`? Couldn't we use `RedisReceiverMetricsConfig::default()` where `RedisReceiverMetricsConfig::new()` is requested and that's it?

metrics_utils/src/lib.rs
```diff
@@ -1267,3 +1270,98 @@ impl BatchMintPersisterMetricsConfig {
...
+        self.transaction_parse_errors.inc_by(count)
+    }
+    pub fn inc_account_parse_errors_by(&self, count: u64) -> u64 {
+        self.account_parse_errors.inc_by(count)
+    }
+    pub fn register(&self, registry: &mut Registry) {
+        registry.register(
+            "redis_receiver_start_time",
+            "Redis receiver metrics start time",
+            self.start_time.clone(),
```
kstepanovdev, line 1328: if `start_time` is `Copy` type it shouldn't require an explicit `clone()`

metrics_utils/src/lib.rs
armyhaylenko: tbh because this seems to be the code standard in the repo

kstepanovdev: Then it's an odd standard we should get rid of, imo. This probably deserves a dedicated ticket.

metrics_utils/src/lib.rs
kstepanovdev: nvm, I looked for types and it's not `Copy`

kstepanovdev (approved):
LGTM

StanChe (approved):
👍

andrii-kl (approved):

PR #388, opened 2025-01-27
[mtg-1030] Test for pseudo-static asset type being converted from fungible into NFT 

kstepanovdev:
- Added a possibility to do integration tests using Eclipse 
- Added test for the case when fungible tokens are returned while only NFTs are requested
- Added tests for multiple transitions between Fungible and NFT metadata


armyhaylenko (approved):
LBTM

andrii-kl (approved):

PR #390, opened 2025-01-28
feat(api): add Raydium price fetcher cache warmup for symbols

armyhaylenko:


nft_ingester/src/api/service.rs
```diff
@@ -111,6 +111,20 @@ pub async fn start_api(
     }
 
     let addr = SocketAddr::from(([0, 0, 0, 0], port));
+    let token_price_fetcher = Arc::new(RaydiumTokenPriceFetcher::new(
+        "https://api-v3.raydium.io".to_string(),
```
kstepanovdev, line 115: nit: I'd like to see it as a constant

nft_ingester/src/raydium_price_fetcher.rs
```diff
@@ -36,6 +36,43 @@ impl RaydiumTokenPriceFetcher {
         }
     }
 
+    pub async fn warmup(&self) -> Result<(), IngesterError> {
+        #[derive(serde::Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct MintListItem {
```
kstepanovdev, line 42: Why do we want a declaration of structs inside a function? Does it do something useful?

kstepanovdev (approved):

kstepanovdev (approved):
Left nit comments. CI is failing though. lgtm in general

nft_ingester/src/raydium_price_fetcher.rs
armyhaylenko: the struct is single-use and used to deserialize only one particular response for one particular function, to me it looks logical & so much better than `.get("field")` used previously

andrii-kl (approved):
looks good

PR #392, opened 2025-01-29
MTG-1277 Fixed sh script launch for running pg docker (test env)

andrii-kl:


armyhaylenko (approved):
LBTM, good catch here

kstepanovdev (approved):

PR #393, opened 2025-01-29
feat(rpc): add exclusive backoff for signature fetch

armyhaylenko:


backfill_rpc/src/rpc.rs
```diff
@@ -8,26 +17,80 @@ use interface::{
...
+                Ok(SignatureWithSlot {
+                    signature: Signature::from_str(&response.signature)?,
+                    slot: response.slot,
+                })
+            })
+            .collect::<Result<Vec<_>, UsecaseError>>()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_with_sender_and_constant_backoff(
```
kstepanovdev, line 88: Can't we just inline the function in the place where it is needed? At first glance, placing it in a trait and hiding it behind a feature flag doesn't look appropriate.

kstepanovdev (changes requested):
Looks super good, however, I have doubts regarding test methods in traits.

backfill_rpc/src/rpc.rs
armyhaylenko: even though it was in a struct, i agree

backfill_rpc/src/rpc.rs
```diff
@@ -173,53 +201,194 @@ impl FinalizedSlotGetter for BackfillRPC {
...
+            )
+        });
+
+        let (addr, server) = warp::serve(route).bind_ephemeral(([127, 0, 0, 1], 0));
+        let _server_handle = tokio::spawn(server);
+
+        let client =
+            RpcClient::new_sender(TooManyRequestsRpcSender(addr.to_string()), Default::default());
+
+        let client = BackfillRPC { client: BackoffRpcClient::new_test(client) };
```
kstepanovdev, line 348:
```suggestion
        let client = BackfillRPC {
            client: {
                let backoff = ExponentialBackoff::from_millis(1).factor(1000);
                BackoffRpcClient {
                    client: Arc::new(client),
                    backoff,
                    is_in_backoff: Arc::new(AtomicBool::new(false)),
                    backoff_finish_notify: Arc::new(Notify::new()),
                }
            },
        };
```

kstepanovdev: instead of using `new_test` method as a part of the trait for test purposes only

kstepanovdev: If it will not be reusable it should be enough. Although if we plan doing this occasionally, then the trait might be ok

armyhaylenko: it's not in the trait really, only in the struct impl, and i've pushed a test that uses the `new_test` function too, so i'll push for keeping it

kstepanovdev (approved):

StanChe (approved):
Great, thank you! The only concern is the error propagation and checking on string

backfill_rpc/src/rpc.rs
```diff
@@ -173,53 +201,276 @@ impl FinalizedSlotGetter for BackfillRPC {
...
+    ) -> Result<R, ClientError> {
+        if self.is_in_backoff.load(Ordering::Relaxed) {
+            self.backoff_finish_notify.notified().await;
+        }
+        let result =
+            RetryIf::spawn(self.backoff.clone().take(MAX_RPC_RETRIES), call, |e: &ClientError| {
+                let kind = e.kind();
+                match kind {
+                    ClientErrorKind::RpcError(RpcError::RpcRequestError(s)) => {
+                        let retry = s.contains("429 Too Many Requests");
```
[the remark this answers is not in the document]
StanChe: Clarified offline, all good.

PR #410, opened 2025-02-10
[MTG-1301] Fix Docker workflow permissions

StanChe:
This PR adds required permissions for GitHub Container Registry access in the Docker workflow.

## Changes
- Add permissions block to docker.yml workflow
- Grant read access to repository contents
- Grant write access to packages for GitHub Container Registry
 
This should fix the 'unauthorized: access token has insufficient scopes' error during Docker image pushes.

Closes #MTG-1301

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,6 +8,11 @@ on:
     branches: [develop]
     tags: ["v*"]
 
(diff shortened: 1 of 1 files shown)


armyhaylenko (approved):
LRFBTM

kstepanovdev (approved):

PR #411, opened 2025-02-10
fix(api): reorder `cursor` and `options` params in api requests

armyhaylenko:
Reorder `cursor` and `options` positioning in RPC request types to align with other das-api implementations in case the user passes RPC params as array.


StanChe (commented):
Can we have a test please

StanChe (commented):
Nice, once CI is green we should merge

kstepanovdev (approved):

StanChe (approved):

PR #416, opened 2025-02-12
[OBEZS] Fix Docker tag generation for branch and tag builds

obezsmertnyi:
- Separate metadata generation for tag and branch builds to prevent invalid tags.
- Ensure correct tagging format for `latest`, `branch-latest`, and `semver`.
- Improve version extraction logic.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,7 +41,11 @@ jobs:
             echo "VERSION_INFO<<EOF" >> "$GITHUB_ENV"
             ./docker/version.sh | sed -e '$a\' >> "$GITHUB_ENV"
             echo "EOF" >> "$GITHUB_ENV"
-            echo "version=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+            ver=$( [[ "$GITHUB_REF" == refs/tags/* ]] \
+                  && echo "${GITHUB_REF#refs/tags/}" \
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -99,25 +103,34 @@ jobs:
...
         with:
           images: |
             ghcr.io/${{ github.repository_owner }}/aura-${{ matrix.binary-name }}
           tags: |
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
-            type=ref,event=branch
-            type=sha
+            type=semver,pattern=v{{version}}
```
armyhaylenko, line 114:
```suggestion
            type=semver,pattern={{version}}
```
nit: i suggest that we must tag our code with tags of `vMAJ.MIN.PATH` and not just `MAJ.MIN.PATCH`, and not rely on the workflow to add the letter "v".

armyhaylenko (approved):
generally LBTM, one comment about versioning

.github/workflows/docker.yml
obezsmertnyi: {{version}} returns only the numeric value (e.g., 1.2.3) as documented [here](https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver).

armyhaylenko (approved):

PR #421, opened 2025-02-14
fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2025-0004

Also updated yanked crates `futures-util`, `spl-discriminator`, `spl-pod`, `spl-program-error`, `spl-tlv-account-resolution` and `spl-type-length-value`.


StanChe:
Thanks for the PR. The vulnerable openssl is updated, but a some crates versions were rolled back to up to 2 major versions. Do you have any insights on that change?

rex4539:
The occur automatically on `cargo update` of specific crates.

rex4539:
Closing due to no interest.

PR #423, opened 2025-02-17
feat(api): rework payload parsing

armyhaylenko:
Two things are changed in this PR:
1) V1 payload parsing is prioritized over V0 as for V0 we do not parse `displayOptions` at all.
2) `showUnverifiedCollections` now defaults to `false`.

### DIFF
--- a/entities/src/api_req_params.rs
+++ b/entities/src/api_req_params.rs
@@ -51,14 +51,10 @@ pub enum AssetSortDirection {
     Desc,
 }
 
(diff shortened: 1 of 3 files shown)


StanChe (approved):
amazing, thank you!
As a side note - gitflow's convention for the feature branch name is `feature/...` and not `feat/`. Soon the GitHub rules will be hardened, not sure that'll affect the feature naming, but lets stick to 1 convention.

kstepanovdev (approved):

PR #437, opened 2025-03-03
feat: improve the deployment process

andrii-kl:
- added the ability to select docker image when starting docker compose


docker-compose.yaml
```diff
@@ -1,5 +1,6 @@
 services:
   ingester:
+    image: ghcr.io/metaplex-foundation/aura-ingester:${DOCKER_IMAGE}
```
StanChe, line 3: shouldn't we also drop build section below?

armyhaylenko (approved):
LBTM

PR #444, opened 2025-03-05
[mtg-1380] fix(api):  getAssetsByCreator doesn't return empty fields

andrii-kl:
- Return empty string for all collection_metadata fields if we have no values

### DIFF
--- a/nft_ingester/src/api/dapi/rpc_asset_convertors.rs
+++ b/nft_ingester/src/api/dapi/rpc_asset_convertors.rs
@@ -238,26 +238,20 @@ fn extract_collection_metadata(
     let link_fields = vec!["name", "symbol"];
     for name in link_fields {
         let value = safe_select(chain_data_selector, format!("$.{}", name).as_str());
-        match value {
-            Some(symbol) => {
-                meta.set_item(name, symbol.clone());
-            },
-            None => {
-                meta.set_item(name, empty_string_value.clone());
-            },
+        if let Some(symbol) = value {
+            meta.set_item(name, symbol.clone());
+        } else {
+            meta.set_item(name, empty_string_value.clone());
         }
     }
 
     let link_fields = vec!["image", "external_url", "description"];
     for name in link_fields {
         let value = safe_select(selector, format!("$.{}", name).as_str());
-        match value {
-            Some(symbol) => {
-                meta.set_item(name, symbol.clone());
-            },
-            None => {
-                meta.set_item(name, empty_string_value.clone());
-            },
+        if let Some(symbol) = value {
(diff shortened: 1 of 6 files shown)


nft_ingester/src/api/dapi/rpc_asset_convertors.rs
```diff
@@ -234,16 +236,29 @@ fn extract_collection_metadata(
...
+    for name in link_fields {
+        let value = safe_select(chain_data_selector, format!("$.{}", name).as_str());
+        match value {
+            Some(symbol) => {
+                meta.set_item(name, symbol.clone());
+            },
+            None => {
+                meta.set_item(name, empty_string_value.clone());
+            },
+        }
```
kstepanovdev, line 249:
```suggestion
         if let Some(symbol) = value {
            meta.set_item(name, symbol.clone());
        } else {
            meta.set_item(name, empty_string_value.clone());
        }
```

nft_ingester/src/api/dapi/rpc_asset_convertors.rs
```diff
@@ -234,16 +236,29 @@ fn extract_collection_metadata(
...
     }
 
     let link_fields = vec!["image", "external_url", "description"];
-    for f in link_fields {
-        let l = safe_select(selector, format!("$.{}", f).as_str());
-        if let Some(l) = l {
-            meta.set_item(f, l.to_owned());
+    for name in link_fields {
+        let value = safe_select(selector, format!("$.{}", name).as_str());
+        match value {
```
kstepanovdev, line 254: The same as above

armyhaylenko (approved):
LBTM

nft_ingester/src/api/dapi/rpc_asset_convertors.rs
```diff
@@ -234,16 +235,29 @@ fn extract_collection_metadata(
...
-        meta.set_item("symbol", symbol.clone());
+    let link_fields = vec!["name", "symbol"];
+    for name in link_fields {
+        let value = safe_select(chain_data_selector, format!("$.{}", name).as_str());
+        match value {
+            Some(symbol) => {
+                meta.set_item(name, symbol.clone());
+            },
+            None => {
+                meta.set_item(name, empty_string_value.clone());
```
kstepanovdev, line 246: `"".into()`  
Won't do the trick? I don't thing we need to create an empty string as a dedicated variable.

andrii-kl: Good trick)

kstepanovdev (approved):
lgtm

PR #446, opened 2025-03-07
fix(rocksdb): prioritize slot ordering over version in comparison logic [MTG-1415]

StanChe:
Change the comparison order in macro to check
slot_updated values first before checking update_version. This ensures account-based assets are properly indexed by slot-first ordering, which better represents the correct temporal sequence of asset updates.

Closes MTG-1415


armyhaylenko (approved):

n00m4d (approved):
Was there an issue that we got account update with higher slot but same or lower write version?
And we still using one Solana node, right? So there couldn't be a situation that we parsed accounts snapshot from one Solana node and then start to listen Geyser updates from the another one

kstepanovdev (approved):

PR #448, opened 2025-03-10
feat(nft_ingester): read snapshot from a backfill stream

armyhaylenko:


andrii-kl (approved):

n00m4d (approved):

StanChe (approved):
Can we also have a different batch size? For backfill that may be several thousands

StanChe (approved):

PR #449, opened 2025-03-11
[MTG-1213] feat(api): implement getNftEditions API function

andrii-kl:
- Implemented new structure to store NFT parent relationship (editions)  information in RocksDB. (store/update)
- Integrated the function with the existing infrastructure
- Updated relevant tests to cover the new functionality
- Refactored related code for better readability and maintainability


entities/src/api_req_params.rs
```diff
@@ -28,6 +28,37 @@ pub struct Pagination {
...
+            pagination.before.is_none() && pagination.after.is_none() && pagination.page.is_none();
+
+        // if cursor is passed use it as 'after' parameter
+        let after = {
+            if cursor_enabled {
+                pagination.cursor
+            } else {
+                pagination.after
+            }
+        };
```
armyhaylenko, line 51:
```suggestion
        let after = if cursor_enabled {
                pagination.cursor
            } else {
                pagination.after
            };
```
but make sure to run the formatting

nft_ingester/src/api/dapi/response.rs
```diff
@@ -75,6 +79,68 @@ impl From<AssetSignature> for SignatureItem {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub page: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub before: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub after: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cursor: Option<String>,
+}
+
```
armyhaylenko, line 143: i strongly suggest removing the "Model" suffix, and if required, specify fully-qualified paths for rocksdb types. worst case, the "Response" suffix is better than "Model"

rocks-db/src/clients/asset_client.rs
```diff
@@ -292,10 +301,11 @@ impl Storage {
                 .get_editions(
                     assets_data
                         .values()
-                        .filter_map(|a: &crate::asset::AssetCompleteDetails| {
+                        .filter_map(|a: &AssetCompleteDetails| {
                             a.static_details.as_ref().map(|s| s.edition_address)
                         })
                         .flatten()
+                        .dedup()
```
armyhaylenko, line 308: are we sure that this data is sorted when we dedup it? cause it will only remove consecutive equal elements

rocks-db/src/clients/asset_client.rs
```diff
@@ -314,19 +324,129 @@ impl Storage {
         })
     }
 
+    pub async fn get_master_edition_child_assets(
+        &self,
+        master_edition: Pubkey,
+        pagination: PaginationQuery,
+    ) -> Result<Vec<TokenMetadataEditionParentIndex>> {
+        let limit = pagination.limit.unwrap_or(100) as usize;
+        let mut revert = false;
```
armyhaylenko, line 333: rename to `reverse`

rocks-db/src/clients/asset_client.rs
```diff
@@ -314,19 +324,129 @@ impl Storage {
...
+
+        let iter = if pagination.after.is_some() {
+            let mut iter = self.token_metadata_edition_parent_index.iter(EditionIndexKey {
+                pub_key: master_edition,
+                edition: pagination
+                    .after
+                    .unwrap()
+                    .parse::<u64>()
+                    .expect("failed to parse edition after key"),
```
armyhaylenko, line 342: i don't think panicking in this case is a good idea

rocks-db/src/clients/asset_client.rs
```diff
@@ -314,19 +324,129 @@ impl Storage {
...
+            iter.next();
+            iter
+        } else if pagination.before.is_some() {
+            let mut iter = self.token_metadata_edition_parent_index.iter_reverse(EditionIndexKey {
+                pub_key: master_edition,
+                edition: pagination
+                    .before
+                    .unwrap()
+                    .parse::<u64>()
+                    .expect("failed to parse edition before key"),
```
armyhaylenko, line 354: no need to panic

rocks-db/src/columns/editions.rs
```diff
@@ -33,6 +35,35 @@ impl TypedColumn for TokenMetadataEdition {
...
+        Ok(EditionIndexKey { pub_key, edition })
+    }
+
+    fn decode(bytes: &[u8]) -> Result<Self::ValueType> {
+        serde_cbor::from_slice(bytes).map_err(|e| StorageError::Common(e.to_string()))
+    }
+
+    fn encode(v: &Self::ValueType) -> Result<Vec<u8>> {
+        serde_cbor::to_vec(&v).map_err(|e| StorageError::Common(e.to_string()))
+    }
```
armyhaylenko, line 64: why do we use CBOR as opposed to bincode?

rocks-db/src/columns/editions.rs
```diff
@@ -78,3 +109,39 @@ pub fn merge_token_metadata_edition(
...
+        }
+    }
+
+    for op in operands {
+        match serde_cbor::from_slice::<TokenMetadataEditionParentIndex>(op) {
+            Ok(new_val) => {
+                if new_val.write_version > write_version || result.is_empty() {
+                    write_version = new_val.write_version;
+                    result = op.to_vec();
+                }
```
armyhaylenko, line 138: @StanChe as with MPL-core accounts, do we need to account for slot_updated here?

rocks-db/src/clients/asset_client.rs
andrii-kl: Yes, it is not necessary here. Above we use hashmap for it.

entities/src/api_req_params.rs
```diff
@@ -150,6 +181,17 @@ pub struct GetAssetsByOwner {
     pub cursor: Option<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct GetNftEditions {
+    pub mint: String,
```
kstepanovdev, line 187: Should not it be `Pubkey`? Or serde cannot automatically convert this type?

entities/src/api_req_params.rs
```diff
@@ -150,6 +181,17 @@ pub struct GetAssetsByOwner {
     pub cursor: Option<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
```
kstepanovdev, line 185: redundant fields prevent de/serialization. is it expected behaviour?

nft_ingester/src/api/api_impl.rs
```diff
@@ -382,6 +386,35 @@ where
...
+
+        let pagination = Pagination {
+            limit: payload.limit.unwrap_or(DEFAULT_LIMIT as u32).into(),
+            page: payload.page,
+            before: payload.before,
+            after: payload.after,
+            cursor: payload.cursor,
+        };
+        Self::validate_basic_pagination(&pagination, self.max_page_limit)?;
+        let mint_address = Pubkey::from_str(&payload.mint)
```
kstepanovdev, line 402: can't this parameter be deserialized by `serde` as I wrote above?

nft_ingester/src/api/api_impl.rs
```diff
@@ -382,6 +386,35 @@ where
...
+            .map_err(|_| DasApiError::Validation("Invalid mint address".to_string()))?;
+
+        let master_asset_editions_info = self
+            .rocks_db
+            .get_master_edition_child_assets_info(
+                mint_address,
+                PaginationQuery::build(pagination.clone()).unwrap(),
```
kstepanovdev, line 409: is this unwrap necessary? seems like we can build the builder with proper error handling.

nft_ingester/src/api/builder.rs
```diff
@@ -98,6 +99,21 @@ impl RpcApiBuilder {
         });
         module.add_alias("getAssetsByCreator", "get_assets_by_creator");
 
+        let cloned_api = api.clone();
+        module.add_method("get_nft_editions", move |rpc_params: Params| {
+            let api = cloned_api.clone();
```
kstepanovdev, line 104: we already did a clone for this variable, seems like a redundant clone. for some reason rust-analyze is not working, so I can confirm this. (all the cases in that impl use the same pattern, which may be redundant)

nft_ingester/src/api/dapi/response.rs
```diff
@@ -75,6 +79,68 @@ impl From<AssetSignature> for SignatureItem {
...
+impl From<(MasterAssetEditionsInfo, Pagination)> for MasterAssetEditionsInfoModel {
+    fn from((value, pagination): (MasterAssetEditionsInfo, Pagination)) -> Self {
+        Self {
+            master_edition_address: value.master_edition_address.to_string(),
+            supply: value.supply,
+            max_supply: value.max_supply,
+            editions: value.editions.iter().map(AssetEditionInfoModel::from).collect(),
+            pagination: PaginationModel {
+                total: value.editions.len() as u32,
```
kstepanovdev, line 118: are 100% sure such a conversion is safe?

rocks-db/src/batch_savers.rs
```diff
@@ -138,8 +141,24 @@ impl BatchSaveStorage {
 
     pub fn store_edition(&mut self, key: Pubkey, edition: &TokenMetadataEdition) -> Result<()> {
         self.storage.token_metadata_edition_cbor.merge_with_batch(&mut self.batch, key, edition)?;
+
+        if let TokenMetadataEdition::EditionV1(edition_v1) = edition {
+            let edition_index = TokenMetadataEditionParentIndex {
```
kstepanovdev, line 146: Looks like a good place for implementing `From` trait.

rocks-db/src/clients/asset_client.rs
```diff
@@ -314,19 +324,129 @@ impl Storage {
...
+                // iter.skip(n) changes the iterator type, making it incompatible with RocksDB.
+                for _ in 0..skip {
+                    iter.next();
+                }
+            }
+
+            iter
+        };
+
+        let mut asset_keys: Vec<TokenMetadataEditionParentIndex> = Vec::with_capacity(limit);
```
kstepanovdev, line 379: If you already know the size, maybe is it better not to use heap-allocated structures? Seems like an array could be used.

rocks-db/src/clients/asset_client.rs
```diff
@@ -314,19 +324,129 @@ impl Storage {
...
+        Ok(MasterAssetEditionsInfo {
+            master_edition_address: Storage::find_edition_address(&master_edition),
+            supply: master_edition_metadata.supply,
+            max_supply: master_edition_metadata.max_supply,
+            editions: asset_edition_info_list,
+        })
+    }
+
+    pub fn find_edition_address(mint: &Pubkey) -> Pubkey {
+        let (edition_address, _) = Pubkey::find_program_address(
```
kstepanovdev, line 430: As long as you do not need seeds, [create\_program\_address](https://docs.rs/solana-sdk/latest/solana_sdk/pubkey/struct.Pubkey.html#method.create_program_address) is the better option because it's more resource effective.

rocks-db/src/columns/editions.rs
kstepanovdev: Indeed. afaik Oleksandr got rid of `Cbor` in favour of `bincode`.

entities/src/api_req_params.rs
andrii-kl: in other places we also use string

entities/src/api_req_params.rs
andrii-kl: Yes, it's a common practise in working with client request. We use it widely for almost all requests

entities/src/api_req_params.rs
kstepanovdev: like here? 

```
let mint_address = Pubkey::from_str(&payload.mint)
```

entities/src/api_req_params.rs
kstepanovdev: Ok then

nft_ingester/src/api/api_impl.rs
andrii-kl: We use convert/validate functions. `validate_pubkey(key)?;`
I have replaced to use it.

nft_ingester/src/api/api_impl.rs
andrii-kl: Without unwrap we have Result< PaginationQuery > but function expected PaginationQuery

kstepanovdev: You can use `?` on it before the function call and handle the error instead of panicking.

nft_ingester/src/api/api_impl.rs
kstepanovdev: You need to convert the parameter each time you want a `Pubkey` because the plain `string` seems to doesn't make much sense. My suggestion is to use a deserializer by `serde` and pass already freshly-baked, nice and already `Pubkey` and catch errors early instead of propagating them through the code because we are used to using `String` instead of `Pubkey`. The code is yours, though, so the choice is as well.

nft_ingester/src/api/builder.rs
andrii-kl: The two clones are necessary because the closure passed to module.add_method needs to own the api instance. The first api.clone() creates a new Arc instance that is then moved into the closure.

nft_ingester/src/api/dapi/response.rs
armyhaylenko: i think solana just wouldn't allow to store more (in byte value) than u32::MAX of editions

nft_ingester/src/api/builder.rs
andrii-kl: but

nft_ingester/src/api/dapi/response.rs
andrii-kl: replaced on u32::try_from

rocks-db/src/clients/asset_client.rs
andrii-kl: maximum limit is 100
I'm not sure that this will bring any growth but at the same time the code itself will become more complicated. 
because we will have Vec<Option<TokenMetadataEditionParentIndex>> and some other things that required additional checks

rocks-db/src/clients/asset_client.rs
andrii-kl: may break if the key matches a regular Pubkey. 
Therefore, it requires additional checks, I am not sure that it will be possible to somehow save resources

rocks-db/src/columns/editions.rs
andrii-kl: I took the current solution that was used in the editions.rs

andrii-kl: We still use cbor for editions, raw black and parameters

rocks-db/src/columns/editions.rs
andrii-kl: Edition entity doesn't have slot information. I implemented the merge function in the same way as for Edition.
If we need check by slots, then this means that the current functionality for saving Edition works incorrectly and requires deeper refactoring.

entities/src/api_req_params.rs
andrii-kl: GetAssetsByOwner, GetAsset, GetAssetBatch, GetAssetProof, GetAssetProofBatch, GetAssetsByCreator, GetAssetsByAuthority, GetGrouping and others in this file. 

I wouldn't mind making this improvement, but it would require changing the current approach as a whole, which would result in a big change.
Perhaps it would be worth making a separate task to make the changes for all objects, and then checking everything after that.

nft_ingester/src/api/api_impl.rs
andrii-kl: Yes, I agree with your point. My latest impl validate `Publkey`  almost immediately and sends an throw an error if value is not valid. 
`let mint_address = validate_pubkey(payload.mint)?;`

armyhaylenko (approved):
LBTM

entities/src/api_req_params.rs
andrii-kl: https://linear.app/mplx/issue/MTG-1441/implement-a-custom-serde-serializerdeserializer-to-automatically

nft_ingester/src/api/api_impl.rs
andrii-kl: https://linear.app/mplx/issue/MTG-1441/implement-a-custom-serde-serializerdeserializer-to-automatically

kstepanovdev (approved):

PR #451, opened 2025-03-13
fix(api): reorder `SearchAssets` fields to comply with V0 repr

armyhaylenko:
Move the `token_type` field in `SeachAssets` to the end of the struct in order to not break pagination on V0 representation (positional JSON case).

This is a **breaking change** for users of positional JSON of V1 variant, but positional JSON seems to not be publicly used outside of umi.


StanChe (approved):
Can we have an integration test with positional json created by uni client showcasing the correct behaviour?

StanChe (approved):
Great, thanks!

kstepanovdev (approved):

PR #452, opened 2025-03-14
[mtg-1433] feat(api): expand health check functionality and improve structure

andrii-kl:
- Added information about the node and application version to the health check
- Improved the structure to make it more flexible and extensible
- Updated relevant tests to cover the new functionality


nft_ingester/src/api/dapi/response.rs
```diff
@@ -23,6 +23,30 @@ pub struct GetGroupingResponse {
...
+    pub node_name: Option<String>,
+    pub checks: Vec<Check>,
+}
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+pub struct Check {
+    pub status: Status,
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
```
kstepanovdev, line 40: Can't this field be a part of status? I do not think we need a dedicated description.

nft_ingester/src/api/api_impl.rs
```diff
@@ -124,12 +127,32 @@ where
...
+                }
+            },
+        };
+
+        let response = HealthCheckResponse {
+            status,
+            app_version: self.health_check_info.app_version.clone(),
+            node_name: self.health_check_info.node_name.clone(),
+            checks: vec![check],
+        };
```
kstepanovdev, line 151:
```
        // I would do sth like this
        
        let label = "check_health";
        self.metrics.inc_requests(label);
        let latency_timer = Instant::now();
        let mut check = Check::new("PostgresDB");

        if self.pg_client.check_health().await.is_err() {
            check.status = Status::Unhealthy(DasApiError::InternalDbError.to_string());
        }

        let response = HealthCheckResponse {
            status: check.status.clone(),
            app_version: self.health_check_info.app_version.clone(),
            node_name: self.health_check_info.node_name.clone(),
            checks: vec![check],
        };

        self.metrics.set_latency(label, latency_timer.elapsed().as_millis() as f64);

        Ok(json!(response))
```

kstepanovdev (approved):
It looks fine in general. Nevertheless, I reckon response structures like `Check` and `HealthCheckResponse` might be slightly improved.

nft_ingester/src/api/dapi/response.rs
andrii-kl: I added this field down for further expansion.
When we have a check for several services and they can have different errors or problems, it will be easier to read the status from each of them.

nft_ingester/src/api/api_impl.rs
andrii-kl: slightly changed current implementation, but left the same idea that we can have several errors and several checks for different services.
 
Like here for example:
```
{
  "status": “Ok”,
  "app_version": "1.2.3",
  "node_name": “prod_node-1”,
  "checks": [
    {
      "name": “PostgresDB”,
      "status": “Ok”,
      "details": null
    },
    {
      "name": “Redis”,
      "status": “Ok”,
      "details": null
    },
    {
      "name": “Rpc”,
      "status": “Ok”,
      "details": null
    },
  ]
}
```

StanChe (approved):
Great job, thank you

PR #456, opened 2025-03-18
fix(api): apply cursor pagination to fungible tokens

armyhaylenko:
Apply curor pagination to both fungible and non-fungible tokens. Previously, the `UNION` we performed always included potentially duplicate fungible tokens, as they were not filtered by the cursor clause.

### DIFF
--- a/nft_ingester/tests/api_tests.rs
+++ b/nft_ingester/tests/api_tests.rs
@@ -4109,6 +4109,419 @@ mod tests {
         assert_eq!(res.items[1].clone().interface, Interface::V1NFT);
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_fungible_and_non_fungible_cursor_pagination() {
+        let cnt = 100;
+        let cli = Cli::default();
+        let (env, _) = setup::TestEnvironment::create_noise(&cli, cnt, 100).await;
+
+        let synchronizer = nft_ingester::index_synchronizer::Synchronizer::new(
+            env.rocks_env.storage.clone(),
+            env.pg_env.client.clone(),
+            200_000,
+            "".to_string(),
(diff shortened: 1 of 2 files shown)


nft_ingester/tests/api_tests.rs
```diff
@@ -4109,6 +4109,419 @@ mod tests {
...
+                show_unverified_collections: true,
+                ..Default::default()
+            }),
+            token_type: Some(TokenType::All),
+            ..Default::default()
+        };
+        let res = api.search_assets(payload).await.unwrap();
+        let res: AssetList = serde_json::from_value(res).unwrap();
+
+        // the default direction of sorting is descending by default.
```
StanChe, line 4470:
```suggestion
        // the default direction of sorting is descending.
```

nft_ingester/tests/api_tests.rs
```diff
@@ -4109,6 +4109,419 @@ mod tests {
...
+            env.pg_env.client.clone(),
+            200_000,
+            "".to_string(),
+            Arc::new(SynchronizerMetricsConfig::new()),
+            1,
+        );
+
+        // create two pseudo-fungible tokens (they are in fact NFTs)
+        let fungible_token_mint1 = Pubkey::new_unique();
+        let fungible_token_mint2 = Pubkey::new_unique();
```
StanChe, line 4129:
```suggestion
        let nft_token_mint1 = Pubkey::new_unique();
        let nft_token_mint2 = Pubkey::new_unique();
```

StanChe (approved):
Great, thanks for fixing it. Some clarifications in the comments. Given our cursor is also working with sorting by updated/created it might make sense to add a test case for it as well - that's not required atm.

andrii-kl (approved):

PR #465, opened 2025-03-31
feat(slot_persister): add periodic missed slot fetch

armyhaylenko:


nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -30,8 +41,15 @@ const MAX_BATCH_RETRIES: usize = 5;
...
-
-#[derive(Parser, Debug)]
+/// How often we perform the backfill check (6 hours by default)
+const DEFAULT_SLOT_BACKFILL_INTERVAL_SECS: u64 = 60 * 60 * 6;
+/// The gap around the checkpoint. If, for example, the last checkpointed slot is
+/// 200_000_000, we will verify slots from 200_000_000 to 201_000_000 (or the last available one).
+/// Then, if we find all the missing slots in this range, we will advance the checkpoint to the next
+/// inconsistent position.
+const DEFAULT_SLOT_BACKFILL_OFFSET: u64 = 1_000_000;
+
```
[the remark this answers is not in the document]
armyhaylenko: it's not very large in fact.

rocks-db/src/columns/raw_block.rs
```diff
@@ -46,6 +47,58 @@ impl TypedColumn for RawBlock {
     }
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct MissedSlotsIdx;
+
+impl TypedColumn for MissedSlotsIdx {
+    type KeyType = (u64, u64);
```
StanChe, line 57: It'll be great to explain that those are concatenation of sequence and slot number

rocks-db/src/columns/parameters.rs
```diff
@@ -8,6 +8,7 @@ pub enum Parameter {
     LastBackfilledSlot,
     LastFetchedSlot,
     TopSeenSlot,
+    LastProcessedMissedSlot,
```
StanChe:
```suggestion
    LastProcessedMissedSlotKey,
```

.env.example
```diff
@@ -39,6 +39,12 @@ ROCKS_SECONDARY_SLOTS_DB_PATH=/path/to/secondary/ingester-slots
 BACKFILLER_SOURCE_MODE=bigtable
 BIG_TABLE_CONFIG='{"creds": "/usr/src/app/creds.json", "timeout": 1000}'
 
+# Slot persister
+SLOTS_DB_PRIMARY_PATH="/path/to/slots/data"
```
StanChe: Why are we not using ROCKS_SLOTS_DB_PATH?

nft_ingester/src/backfiller.rs
```diff
@@ -213,6 +233,64 @@ where
...
+            error!("Error processing slot {}: {}", slot, e);
+        }
+        db.put_parameter(Parameter::LastProcessedMissedSlot, (seq, slot)).await?;
+
+        if let Some(block_time) = block_time {
+            let dur = time::SystemTime::now()
+                .duration_since(time::UNIX_EPOCH + Duration::from_secs(block_time))
+                .unwrap_or_default()
+                .as_millis() as f64;
+            metrics.set_slot_delay_time("missed_slot_backfilled", dur);
```
StanChe: slot_delay is a histogram that has not that many buckets. so this metric makes sense only if we account for the sum/count

nft_ingester/src/bin/ingester/main.rs
```diff
@@ -544,23 +544,48 @@ pub async fn main() -> Result<(), IngesterError> {
...
-                tx_ingester.clone(),
-                primary_rocks_storage.clone(),
-                metrics_state.backfiller_metrics.clone(),
-            ));
-            let db = primary_rocks_storage.clone();
-            let metrics: Arc<BackfillerMetricsConfig> = metrics_state.backfiller_metrics.clone();
-            let slot_db = slot_db.clone();
             usecase::executor::spawn({
                 let cancellation_token = cancellation_token.child_token();
+                let consumer = Arc::new(DirectBlockParser::new(
```
StanChe: and we wanted to have 2 distinct block parsers here for?

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +559,261 @@ async fn process_slots(
...
+        );
+
+        // Collect slots from last persisted slot down to 0
+        in_mem_dumper.clear().await;
+
+        // Start slot collection
+        let _ = slots_collector
+            .collect_slots(
+                &BUBBLEGUM_PROGRAM_ID,
+                lower_slot_boundary + slot_backfill_offset,
```
StanChe, line 638: A unlikely but possible situation on a network that has really dispersed transactions and may have 0 slots with bubblegum transactions in a continuous range over the slot_backfill_offset. In this case the method will not be able to get any slots and will potentially hang infinitely. This might apply even to the solana mainnet if applied to some early days of bubblegum.

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +559,261 @@ async fn process_slots(
...
+
+        // Verification loop
+        while let Some(slot) = next_slot {
+            if cancellation_token.is_cancelled() {
+                info!("Shutdown signal received, stopping...");
+                return None;
+            }
+
+            if let Some(db_slot) = current_db_slot {
+                if slot < lower_slot_boundary {
```
StanChe: this check doesn't look right to me. we're comparing our received slot with the lower bound (which should never be possible due to `collected_slots.retain(|s| *s > lower_slot_boundary);` above 
Did you mean to compare the db key? which also should not happen

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +559,261 @@ async fn process_slots(
...
+
+        let mut current_db_slot = if db_iter.valid() {
+            if let Some(key_bytes) = db_iter.key() {
+                RawBlock::decode_key(key_bytes.to_vec()).ok()
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
StanChe:
```suggestion
        let mut current_db_slot = db_iter.key().map(|key_bytes| RawBlock::decode_key(key_bytes.to_vec()).ok());
```
should do the same. The .key() method returns an Option and it checks the validity of the iterator as well. should be more readable this way.

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +559,261 @@ async fn process_slots(
...
+        let collector = slots_collector.clone();
+        let (last_checkpoint, next_checkpoint, missed_slots) = get_missed_slots(
+            dumper,
+            collector,
+            db,
+            slot_backfill_offset,
+            cancellation_token.child_token(),
+        )
+        .await
+        .expect("retrieval of missed slots must join");
```
StanChe: is this a way to terminate the loop on receiving the cancelation token? As none is only returned on cancellation token termination

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +559,261 @@ async fn process_slots(
...
+            .delete(last_checkpoint)
+            .expect("must delete the previous checkpoint after slot processing");
+        debug!(%next_checkpoint, "Setting the new checkpoint value");
+        target_db
+            .slot_consistency_checkpoint
+            .put(next_checkpoint, SlotConsistencyCheckpoint)
+            .expect("must update the new checkpoint after the slots are saved into storage");
+        tokio::select! {
+            _ = cancellation_token.cancelled() => { break; }
+            _ = tokio::time::sleep(Duration::from_secs(args.slot_backfill_interval)) => {
```
StanChe, line 799: Are we ok with the initial check taking 50 days? 200M slots of range with 1M every 6 hours will take long

.env.example
armyhaylenko: the env value was explicitly set to `SLOTS_DB_PRIMARY_PATH` in slot persister binary, happy to unify it and use `ROCKS_SLOTS_DB_PATH`

nft_ingester/src/backfiller.rs
armyhaylenko: i'll remove this metric. i only included it because this code was shamelessly copied from `backfill_slots`.

nft_ingester/src/bin/ingester/main.rs
armyhaylenko: we didn't, i'll make it use one if there are no security issues

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: in my experience, it didn't hang. if we didn't collect any slots in this range we just move the checkpoint forward.

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: i agree, this was an initial check, but later i added the `retain`. will remove

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: agree, this is the result of a shameless copy from slot-checker.

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: this was tested with 1M range every 120 seconds and worked fine; i will lower the limits drastically.

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: great catch, will replace with a match

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -40,11 +62,11 @@ const SLOT_COLLECTION_OFFSET: u64 = 300;
 )]
 struct Args {
     /// Path to the target RocksDB instance with slots
-    #[arg(short, long, env = "SLOTS_DB_PRIMARY_PATH")]
+    #[arg(short, long, env = "ROCKS_SLOTS_DB_PATH")]
```
StanChe, line 70: docker-compose also has the leftovers of
```
    environment:
      SLOTS_DB_PRIMARY_PATH: ${ROCKS_SLOTS_DB_PATH}
```

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +560,244 @@ async fn process_slots(
...
+    /// assert_eq!(slot_count, 50_000u64);
+    ///
+    /// let lower_slot_boundary: u64 = 10_000;
+    /// let higher_slot_boundary = 20_000u64 - MISSED_SLOT_COLLECTION_OFFSET;
+    /// let slot_count =
+    /// higher_slot_boundary.saturating_sub(lower_slot_boundary).min(MAX_SLOTS_TO_ADVANCE);
+    ///
+    /// assert_eq!(slot_count, 7_000u64);
+    /// ```
+    async fn get_missed_slots(
```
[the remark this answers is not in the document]
StanChe: I'll support this. Even recommend moving this logic to the use case layer as the main is growing significantly already. But this one is not a show stopper

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +560,244 @@ async fn process_slots(
...
+                        // Advance both iterators
+                        {
+                            let missing_slots_lock = missing_slots.lock().unwrap();
+                            if missing_slots_lock.is_empty() {
+                                last_valid_key = db_slot;
+                            }
+                        }
+                        next_slot = slots_iter.next();
+                        db_iter.next();
+                        current_db_slot = if db_iter.valid() {
```
StanChe: same comment as before `db_iter.key().and_then(|key_bytes| RawBlock::decode_key(key_bytes.to_vec()).ok());`

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +560,244 @@ async fn process_slots(
...
+                            missing_slots_lock.push(slot);
+                        }
+                        // Advance slots iterator
+                        next_slot = slots_iter.next();
+                    },
+                    Ordering::Greater => {
+                        // slot > db_slot
+                        // Advance RocksDB iterator
+                        db_iter.next();
+                        current_db_slot = if db_iter.valid() {
```
StanChe: and here

nft_ingester/src/bin/slot_persister/main.rs
StanChe: My concern is that if there are no slots we keep last_valid_key = lower_slot_boundary; and will start over the same range the next iteration

StanChe (approved):
Great job, thank you!
Some minor concerns for the potential infinite loop and some optional refactor.

nft_ingester/src/bin/slot_persister/main.rs
```diff
@@ -497,3 +560,232 @@ async fn process_slots(
...
+                return None;
+            }
+
+            if let Some(db_slot) = current_db_slot {
+                match slot.cmp(&db_slot) {
+                    Ordering::Equal => {
+                        // Slot exists in RocksDB
+                        // Advance both iterators
+                        {
+                            let missing_slots_lock = missing_slots.lock().unwrap();
```
StanChe, line 694: why do we need a mutex on missing_slots? this looks to be a perfectly synchronous method in terms of access to it.

StanChe (approved):

nft_ingester/src/bin/slot_persister/main.rs
armyhaylenko: shameless copy of the code from slot-checker is the reason. i'll add a task in our tracker to remove mutexes from both places

